PR #2, opened 2023-10-30
Charles Ren

chuckrenny:
This mid mod was definitely more challenging than the last IC. The ActiveRecord methods felt smooth but now after not coding CRUD, I feel rusty. I'd say this was slightly above average difficulty and it was a good test. Thanks

### DIFF
--- a/app/controllers/movie_actors_controller.rb
+++ b/app/controllers/movie_actors_controller.rb
@@ -1,0 +1,9 @@
+class MovieActorsController < ApplicationController
+  def create
+    @movie = Movie.find(params[:id])
+    actor = Actor.find(params[:actor_id])
+    @movie.actors << actor
+
+    redirect_to "/movies/#{@movie.id}"
+  end
+end
--- a/app/models/actor.rb
+++ b/app/models/actor.rb
@@ -1,0 +1,4 @@
+class Actor < ApplicationRecord
+  has_many :movie_actors
+  has_many :movies, through: :movie_actors
+end
--- a/app/models/actor.rb
+++ b/app/models/actor.rb
@@ -1,4 +1,12 @@
 class Actor < ApplicationRecord
   has_many :movie_actors
   has_many :movies, through: :movie_actors
+
+  def self.ordered_by_age
+    order(:age).pluck(:name)
+  end
+
+  def self.average_age
+    average(:age)
+  end
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -1,3 +1,5 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_many :movie_actors
+  has_many :actors, through: :movie_actors
 end
--- a/app/views/movies/show.html.erb
+++ b/app/views/movies/show.html.erb
@@ -8,3 +8,10 @@
   <p><%= actor.name %></p>
   <p><%= actor.age %></p>
 <% end %>
+
+<h4>Add an Actor Form</h4>
+<%= form_with url: "/movies/#{@movie.id}", method: :post, data: {turbo: false} do |form| %>
+  <%= form.label :actor_id, "Actor ID:" %>
+  <%= form.text_field :actor_id %>
+  <%= form.submit "Add Actor" %>
+<% end %>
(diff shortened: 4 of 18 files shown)


app/controllers/movie_actors_controller.rb
```diff
@@ -0,0 +1,9 @@
+class MovieActorsController < ApplicationController
+  def create
+    @movie = Movie.find(params[:id])
+    actor = Actor.find(params[:actor_id])
+    @movie.actors << actor
+
+    redirect_to "/movies/#{@movie.id}"
```
juliet-e, line 7: Nice controller organization and adherence to REST!

app/models/actor.rb
```diff
@@ -0,0 +1,12 @@
+class Actor < ApplicationRecord
+  has_many :movie_actors
+  has_many :movies, through: :movie_actors
+
+  def self.ordered_by_age
```
juliet-e, line 5: The US actually is looking for this data to be for the actors in a *specific movie* so these calculations would make more sense in the Movie model, where we can call the actor attribute and it will just be a collection of actors associated to that movie:
`actors.order(:age)`

juliet-e: This is true for both the model methods here -- they would make more sense instance methods in Movie. However, you are calling them on `movie.actors` so things will still work, it is just slightly less conventional (but you are not penalized for this, just think about it for next time!)

app/models/movie.rb
```diff
@@ -1,3 +1,5 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_many :movie_actors
+  has_many :actors, through: :movie_actors
```
juliet-e, line 4: Nice M2M set up!

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,17 @@
+<h1><%= @movie.title %> Show Page </h1>
+<p><%=  @movie.creation_year%></p>
+<p><%=  @movie.genre%></p>
+<p><%= @movie.actors.ordered_by_age.join(", ") %></p>
```
juliet-e, line 4: Nice work adding `.join` here - because this is dealing with display logic, this can be called in the view!

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,17 @@
...
+<p>Average Age: <%= @movie.actors.average_age %>
+
+<% @movie.actors.each do |actor| %>
+  <p><%= actor.name %></p>
+  <p><%= actor.age %></p>
+<% end %>
+
+<h4>Add an Actor Form</h4>
+<%= form_with url: "/movies/#{@movie.id}", method: :post, data: {turbo: false} do |form| %>
```
juliet-e, line 13: Looks great as a POST request, but a slightly more RESTful path could be "movies/:id/actors" to indicate that you are creating an association between movies *and actors* rather than just including the actor resource in the path

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  # US 2
+  it "displays the movie attributes and a list of actors ordered by age" do
+    visit "/movies/#{@TS.id}"
+
+    expect(page).to have_content("Toy Story")
+    expect(page).to have_content("2008")
+    expect(page).to have_content("Fictional")
+    expect(page).to have_content("Buzz, Piggy, Woody")
```
juliet-e, line 36: Joining the actors in a comma separated list makes it easier to test order in a feature test! Nice work

spec/features/studios/index_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    @Hulk = @hollywood.movies.create!(title: "Hulk", creation_year: "2010", genre: "Action")
+
+    @Dune = @paramount.movies.create!(title: "Dune", creation_year: "2021", genre: "Action")
+    @HungerGames = @paramount.movies.create!(title: "Hunger Games", creation_year: "2012", genre: "Action")
+  end
+
+  it "displays the studios name, location, and movies and corresponding attributes" do
+    visit "/studios"
+
+    within("#studio-#{@hollywood.id}") do
```
juliet-e, line 18: Nice use of within blocks! You could even do movie within blocks inside this within block! But I know time is tight :)

spec/models/actor_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+      end
+    end
+
+    describe "#average_age" do
+      it "outputs the average age" do
+        expect(Actor.average_age).to eq(9)
+      end
+    end
+  end
+end
```
juliet-e, line 28: Excellent work testing this logic at both the unit (model) and end-to-end (feature spec) level!

PR #3, opened 2023-10-30
Michael Kuhlmeier

mbkuhl:
Smooth. Better grasp on what goes where in MVC

### DIFF
--- a/app/controllers/movie_actors_controller.rb
+++ b/app/controllers/movie_actors_controller.rb
@@ -1,0 +1,12 @@
+class MovieActorsController < ApplicationController
+
+  def new
+    MovieActor.create!(
+      movie_id: params[:id],
+      actor_id: params[:add_actor]
+    )
+    @movie = Movie.find(params[:id])
+    redirect_to "/movies/#{@movie.id}"
+  end
+
+end
--- a/app/models/studio.rb
+++ b/app/models/studio.rb
@@ -2,4 +2,8 @@ class Studio < ApplicationRecord
   validates :name, presence: true
   validates :location, presence: true
   has_many :movies
+
+  def actors
+    Actor.joins(movies: :studio).where("studios.id = ?", id).distinct.pluck(:name)
+  end
 end
--- a/app/models/studio.rb
+++ b/app/models/studio.rb
@@ -1,3 +1,5 @@
 class Studio < ApplicationRecord
+  validates :name, presence: true
+  validates :location, presence: true
   has_many :movies
 end
--- a/app/views/movies/show.html.erb
+++ b/app/views/movies/show.html.erb
@@ -7,9 +7,9 @@
 <% @movie.actors_youngest_to_oldest.each do |actor| %>
   <%= actor %>
 <% end %>
-
+<br><br>
 <%= "Average age of actors: #{@movie.average_age_of_actors}" %>
-
+<br><br>
 <%= form_with url: "/movies/#{@movie.id}", method: :patch, local: :true do |f| %>
   <%= f.text_field :add_actor %>
   <%= f.submit "Add actor" %>
--- a/app/views/movies/show.html.erb
+++ b/app/views/movies/show.html.erb
@@ -9,3 +9,8 @@
 <% end %>
 
 <%= "Average age of actors: #{@movie.average_age_of_actors}" %>
+
+<%= form_with url: "/movies/#{@movie.id}", method: :patch, local: :true do |f| %>
+  <%= f.text_field :add_actor %>
+  <%= f.submit "Add actor" %>
+<% end %>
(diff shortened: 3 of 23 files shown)


app/controllers/movie_actors_controller.rb
```diff
@@ -0,0 +1,12 @@
+class MovieActorsController < ApplicationController
+
+  def new
```
cjsim89, line 3: The "meat" of this method is great - I think it is better called `create` instead of `new` though. In Rails, `new` is understood to be an action that shows a form that _leads to_ a `create` method. Otherwise good!

app/models/studio.rb
```diff
@@ -1,3 +1,9 @@
 class Studio < ApplicationRecord
+  validates :name, presence: true
+  validates :location, presence: true
   has_many :movies
+
+  def actors
+    Actor.joins(movies: :studio).where("studios.id = ?", id).distinct.pluck(:name)
```
cjsim89, line 7: Nice AR! This query is fine, but we can write an even shorter query by relying on associations. Check this out: 

If we add `has_many :actors, through: :movies` to this Studio model, then our query becomes
```ruby
actors.distinct.pluck(:name)
```

The reason this works is because a Movie already `has_many :actors`, so thanks to some transitive logic, a Studio also has many Actors through its Movies. 

Both ways of writing this query works!

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,16 @@
...
+<%= "Genre: #{@movie.genre}" %> <br>
+  <br>
+<h4>Actors, Youngest to Oldest:</h4>
+<% @movie.actors_youngest_to_oldest.each do |actor| %>
+  <%= actor %>
+<% end %>
+<br><br>
+<%= "Average age of actors: #{@movie.average_age_of_actors}" %>
+<br><br>
+<%= form_with url: "/movies/#{@movie.id}", method: :patch, local: :true do |f| %>
```
cjsim89, line 13: The form looks good, but I think instead of `:patch` here we should use `:post`, since we are _creating a relationship_ between this movie and the actor by the ID we provide. `PATCH` and `PUT` are verbs we should use when updating a resource - and if we are updating this movie resource, we would change one of the  values of the `movies` table. Since that's not what we're doing, a different verb should be used, in this case POST.

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+    @movie2 = @studio2.movies.create!(title: "Frozen", creation_year: "2013", genre: "animation")
+    @movie3 = @studio2.movies.create!(title: "Up", creation_year: "2009", genre: "animation")
+    @actor1 = @movie1.actors.create!(name: "Actor 1", age: "20")
+    @actor2 = @movie1.actors.create!(name: "Actor 2", age: "10")
+    @actor3 = @movie1.actors.create!(name: "Actor 3", age: "30")
+    @actor4 = @movie2.actors.create!(name: "Actor 4", age: "40")
+    @actor5 = @movie2.actors.create!(name: "Actor 5", age: "50")
+    @actor6 = @movie3.actors.create!(name: "Actor 6", age: "60")
+
+    @movie1.actors << @actor6
```
cjsim89, line 17: Good test setup overall!

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+      visit "/movies/#{@movie3.id}"
+      expect(page).to have_content("Actors, Youngest to Oldest:")
+      expect(page).to_not have_content("Actor 1")
+      expect(page).to_not have_content("Actor 2")
+      expect(page).to_not have_content("Actor 3")
+      expect(page).to_not have_content("Actor 4")
+      expect(page).to_not have_content("Actor 5")
+      expect(page).to have_content("Actor 6")
+      
+      expect(page).to have_content("Average age of actors: 60")
```
cjsim89, line 76: Really good job testing different outcomes per object!

cjsim89: In general this his probably not needed, I think if you wanted to see that the value can change, you can set up 2 scenarios - that's usually all you'll need.

spec/features/studios/index_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+        end
+      end
+      within "#Disney" do
+        expect(page).to have_content("Movies:")
+        within "#Frozen" do
+          expect(page).to have_content("Frozen")
+          expect(page).to have_content("Year of Creation: 2013")
+          expect(page).to have_content("Genre: animation")
+        end
+        within "#Up" do
```
cjsim89, line 44: Good uses of `within` blocks 👍

cjsim89 (commented):
Great work overall!

PR #4, opened 2023-10-30
Sam T

Sykogst:
I think the IC went ok, I finished everything including the extension.

I was uncertain if adding the relations to the Studio model (has_many :actors, through: :movies) was an okay thing to do or not in able to get a list of unique actors in the studio with the method that I had made.

I also attempted some data validations in the models and was wondering if these are the right way to do them and if it is sufficiently covering everything.

### DIFF
--- a/app/controllers/movie_actors_controller.rb
+++ b/app/controllers/movie_actors_controller.rb
@@ -1,0 +1,9 @@
+class MovieActorsController < ApplicationController
+  def create
+    @actor = Actor.find(params[:actor])
+    @movie = Movie.find(params[:movie_id])
+    ActorMovie.create!(actor: @actor, movie: @movie)
+
+    redirect_to "/movies/#{@movie.id}"
+  end
+end
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -1,0 +1,5 @@
+class MoviesController < ApplicationController
+  def index
+    @movie = Movie.find(params[:id])
+  end
+end
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -1,5 +1,5 @@
 class MoviesController < ApplicationController
-  def index
+  def show
     @movie = Movie.find(params[:id])
   end
 end
--- a/app/controllers/studios_controller.rb
+++ b/app/controllers/studios_controller.rb
@@ -1,0 +1,5 @@
+class StudiosController < ApplicationController
+  def index
+    @studios = Studio.all
+  end
+end
--- a/app/controllers/studios_controller.rb
+++ b/app/controllers/studios_controller.rb
@@ -2,4 +2,8 @@ class StudiosController < ApplicationController
   def index
     @studios = Studio.all
   end
+
+  def show
+    @studio = Studio.find(params[:id])
+  end
 end
--- a/app/models/actor.rb
+++ b/app/models/actor.rb
@@ -1,3 +1,4 @@
 class Actor < ApplicationRecord
   has_many :actor_movies
+  has_many :movies, through: :actor_movies
 end
--- a/app/models/actor.rb
+++ b/app/models/actor.rb
@@ -1,0 +1,3 @@
+class Actor < ApplicationRecord
+  has_many :actor_movies
+end
--- a/app/models/actor.rb
+++ b/app/models/actor.rb
@@ -1,4 +1,7 @@
 class Actor < ApplicationRecord
   has_many :actor_movies
   has_many :movies, through: :actor_movies
+
+  validates :name, presence: true
+  validates :age, presence: true, numericality: true
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -1,3 +1,5 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_many :actor_movies
+  has_many :actors, through: :actor_movies
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -3,6 +3,10 @@ class Movie < ApplicationRecord
   has_many :actor_movies
   has_many :actors, through: :actor_movies
 
+  validates :title, presence: true
+  validates :creation_year, presence: true
+  validates :genre, presence: true
+
   def actors_sorted
     self.actors.order(:age)
   end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -2,4 +2,16 @@ class Movie < ApplicationRecord
   belongs_to :studio
   has_many :actor_movies
   has_many :actors, through: :actor_movies
+
+  def actors_sorted
+    self.actors.order(:age)
+  end
+
+  def actor_age_avg
+    if self.actors.count(:id) == 0
+      0
+    else
+      self.actors.average(:age).round(2)
+    end
+  end
 end
--- a/app/models/studio.rb
+++ b/app/models/studio.rb
@@ -1,3 +1,8 @@
 class Studio < ApplicationRecord
   has_many :movies
+  has_many :actors, through: :movies
+
+  def actors_all
+    self.actors.select("actors.*").distinct
+  end
 end
--- a/app/models/studio.rb
+++ b/app/models/studio.rb
@@ -7,6 +7,6 @@ class Studio < ApplicationRecord
 
 
   def actors_all
-    self.actors.select("actors.*").distinct
+    self.actors.distinct
   end
 end
--- a/app/models/studio.rb
+++ b/app/models/studio.rb
@@ -2,6 +2,10 @@ class Studio < ApplicationRecord
   has_many :movies
   has_many :actors, through: :movies
 
+  validates :name, presence: true
+  validates :location, presence: true
+
+
   def actors_all
     self.actors.select("actors.*").distinct
   end
--- a/app/views/movies/show.html.erb
+++ b/app/views/movies/show.html.erb
@@ -9,3 +9,11 @@
     <%= actor.age %>
   </div>
 <% end %>
+
+<%= form_with url: "/movies/#{@movie.id}/actors", method: :post, data: { turbo: false } do |f| %>
+  <%= f.label :actor %>
+  <%= f.text_field :actor %>
+
+  <%= f.hidden_field :movie_id, value: @movie.id %>
+  <%= f.submit "Add Actor" %>
+<% end %>
--- a/app/views/movies/show.html.erb
+++ b/app/views/movies/show.html.erb
@@ -3,7 +3,7 @@
 <%= @movie.genre %>
 <%= @movie.actor_age_avg %>
 
-<% @employee.actors_sorted.each do |actor| %>
+<% @movie.actors_sorted.each do |actor| %>
   <div id="actor-<%= actor.id %>">
     <%= actor.name %>
     <%= actor.age %>
--- a/app/views/movies/show.html.erb
+++ b/app/views/movies/show.html.erb
@@ -1,0 +1,11 @@
+<%= @movie.title %>
+<%= @movie.creation_year %>
+<%= @movie.genre %>
+<%= @movie.actor_age_avg %>
+
+<% @employee.actors_sorted.each do |actor| %>
+  <div id="actor-<%= actor.id %>">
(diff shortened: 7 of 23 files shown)


app/controllers/movie_actors_controller.rb
```diff
@@ -0,0 +1,9 @@
+class MovieActorsController < ApplicationController
+  def create
+    @actor = Actor.find(params[:actor])
+    @movie = Movie.find(params[:movie_id])
+    ActorMovie.create!(actor: @actor, movie: @movie)
+
+    redirect_to "/movies/#{@movie.id}"
+  end
```
abdulredd, line 8: Perfect. Great use of the join controller to make the connection between an actor and a movie!

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,5 @@
+class MoviesController < ApplicationController
+  def show
+    @movie = Movie.find(params[:id])
+  end
```
abdulredd, line 4: 👍

app/controllers/studios_controller.rb
```diff
@@ -0,0 +1,9 @@
+class StudiosController < ApplicationController
+  def index
+    @studios = Studio.all
+  end
+
+  def show
+    @studio = Studio.find(params[:id])
+  end
+end
```
abdulredd, line 9: 👍

app/models/actor.rb
```diff
@@ -0,0 +1,7 @@
+class Actor < ApplicationRecord
+  has_many :actor_movies
+  has_many :movies, through: :actor_movies
+
+  validates :name, presence: true
+  validates :age, presence: true, numericality: true
+end
```
abdulredd, line 7: Great job adding validations here Sam!

app/models/movie.rb
```diff
@@ -1,3 +1,21 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_many :actor_movies
+  has_many :actors, through: :actor_movies
+
+  validates :title, presence: true
+  validates :creation_year, presence: true
+  validates :genre, presence: true
```
abdulredd, line 8: Did you know there are other ways to write this also so you don't end up repeating yourself?

`validates :title, :creation_year, :genre, presence: true`

and

`validates_presence_of :title, :creation_year, :genre`

app/models/movie.rb
```diff
@@ -1,3 +1,21 @@
...
+    self.actors.order(:age)
+  end
+
+  def actor_age_avg
+    if self.actors.count(:id) == 0
+      0
+    else
+      self.actors.average(:age).round(2)
+    end
+  end
```
abdulredd, line 20: I just want to make this quick note here that depending on who you ask, doing something like a `round` here could be considered presentation logic. I won't mark this as an MVC infraction, but something I wanted to bring up because some rubyist would consider this an infraction because it is technically presentation logic, I am in that camp.

app/models/studio.rb
```diff
@@ -1,3 +1,12 @@
...
   has_many :movies
+  has_many :actors, through: :movies
+
+  validates :name, presence: true
+  validates :location, presence: true
+
+
+  def actors_all
+    self.actors.distinct
+  end
```
abdulredd, line 11: I personally think `actors_unique` would have been a more descriptive name 😉 .

app/models/studio.rb
```diff
@@ -1,3 +1,12 @@
 class Studio < ApplicationRecord
   has_many :movies
+  has_many :actors, through: :movies
```
abdulredd, line 3: Not many people know you can do this, and it took me a while to understand this when I first encountered it. Great job making use of Rails to get to `:actors` through the association that the `Movie` model already has through `:movie_actors`!

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,19 @@
...
+<%= @movie.creation_year %>
+<%= @movie.genre %>
+<%= @movie.actor_age_avg %>
+
+<% @movie.actors_sorted.each do |actor| %>
+  <div id="actor-<%= actor.id %>">
+    <%= actor.name %>
+    <%= actor.age %>
+  </div>
+<% end %>
```
abdulredd, line 11: While as backend devs we are always going to be tasked with writing HTML, since we are creating our own views, you will need to use HTML in your logic. Right now your erb tags are just hanging out all willy-nilly. They should all be enclosed in some sort of HTML tags. For example:

```html.erb
<h1><%= @movie.title %></h1>
<h2><%= @movie.creation_year %><h2>
<h2><%= @movie.genre %><h2>
<p>Average actor age: <%= @movie.actor_age_avg %></p>
```

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,19 @@
...
+
+<%= form_with url: "/movies/#{@movie.id}/actors", method: :post, data: { turbo: false } do |f| %>
+  <%= f.label :actor %>
+  <%= f.text_field :actor %>
+
+  <%= f.hidden_field :movie_id, value: @movie.id %>
+  <%= f.submit "Add Actor" %>
+<% end %>
```
abdulredd, line 19: Yes! This form is perfect and follows restful conventions perfectly! You're using the correct verb (POST) and sending the data to the correct path, great job Sam!

The only thing I will say is that you do not need the hidden field for the movie ID since you are going to be getting the movie ID from the URL. This can be removed.

app/views/studios/index.html.erb
```diff
@@ -0,0 +1,11 @@
...
+<% @studios.each do |studio| %>
+  <div id="studio-<%= studio.id %>">
+    <%= studio.name %>
+    <%= studio.location %>
+    <p>Movies:</p>
+      <% studio.movies.each do |movie| %>
+        <p><%= "#{movie.title}, created: #{movie.creation_year}, genre: #{movie.genre}" %></p>
+      <% end %>
+  </div>
+<% end %>
```
abdulredd, line 11: Great job introducing HTML, but any rendered erb tag should be enclosed in HTML.

app/views/studios/show.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= @studio.name %>
+<%= @studio.location %>
+<% @studio.actors_all.each do |actor| %>
+  <%= actor.name %>
+<% end %>
```
abdulredd, line 5: 🎶 _HTML_ 🎵

config/routes.rb
```diff
@@ -3,4 +3,11 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get '/studios', to: 'studios#index'
+  get '/studios/:id', to: 'studios#show'
+
+  get '/movies/:id', to: 'movies#show'
+
+  post '/movies/:movie_id/actors', to: 'movie_actors#create'
```

db/migrate/20231030151453_create_actor_movies.rb
```diff
@@ -0,0 +1,10 @@
+class CreateActorMovies < ActiveRecord::Migration[7.0]
+  def change
+    create_table :actor_movies do |t|
+      t.references :movie, null: false, foreign_key: true
+      t.references :actor, null: false, foreign_key: true
+
+      t.timestamps
+    end
+  end
```
abdulredd, line 9: Great job creating the joins table.

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+
+      visit "movies/#{@movie_1.id}"
+      expect(page).to have_content(@movie_1.actor_age_avg)
+
+      oldest = find("#actor-#{@actor_2.id}")
+      mid = find("#actor-#{@actor_1.id}")
+      youngest = find("#actor-#{@actor_3.id}")
+
+      expect(youngest).to appear_before(mid)
+      expect(mid).to appear_before(oldest)
```
abdulredd, line 40: I see what you're doing here with the ordering.... but I think you could have gotten away with just doing something like:

```ruby
expect(@actor_3).to appear_before(@actor_1)
expect(@actor_1).to appear_before(@actor_2)
```

This way you don't have to use `find`.

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+
+      expect(current_path).to eq("/movies/#{@movie_1.id}")
+      
+      within "#actor-#{@actor_1.id}" do
+        expect(page).to have_content(@actor_1.name)
+      end
+    end
+  end
+  
+end
```
abdulredd, line 91: These are great tests and I love your use of `within` blocks.

I only have one suggestion for the future to further reduce the ambiguity by using more specific assertions. Here's what I mean by that. This assertion, `expect(page).to have_content(@movie_1.creation_year)` just says that `1995` just needs to show up on the page somewhere, it could be part of a date, a phone number, etc. It doesn't specify that it's the year of the movie at all.

This could be resolved by placing this inside a `within` block or specifying a more specific assertion like: `expect(page).to have_content("Year: #{@movie_1.creation_year}")`. While this may seem like a small thing, it helps reduce ambiguity and helps avoid false positives by testing exactly what and where you intend to test.

spec/features/studios/index_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      end
+
+      within "#studio-#{@studio_2.id}" do
+      expect(page).to have_content("#{@movie_3.title}, created: #{@movie_3.creation_year}, genre: #{@movie_3.genre}")
+      expect(page).to have_content("#{@movie_4.title}, created: #{@movie_4.creation_year}, genre: #{@movie_4.genre}")
+    end
+    end
+  end
+
+end
```
abdulredd, line 44: Great job using `within` blocks here!

spec/features/studios/show_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      visit "studios/#{@studio_1.id}"
+
+      # expect(page).to have_content("Actors: #{@actor_1.name} #{@actor_2.name}")
+      expect(page).to have_content(@actor_1.name, count: 1)
+      expect(page).to have_content(@actor_2.name, count: 1)
+      expect(page).to_not have_content(@actor_3.name)
+    end
+  end
+
+end
```
abdulredd, line 49: You could actually put your `visit` in your `before :each`. This goes for all your tests.

spec/models/actor_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+    it { should have_many :actor_movies }
+    it { should have_many(:movies).through(:actor_movies) }
+  end
+
+  describe "validations" do
+    it { should validate_presence_of(:name) }
+    it { should validate_presence_of(:age) }
+    it { should validate_numericality_of(:age) }
+  end
+end
```
abdulredd, line 14: Perfect validation tests!

spec/models/movie_spec.rb
```diff
@@ -2,6 +2,52 @@
...
+    it { should belong_to :studio }
+    it { should have_many :actor_movies }
+    it { should have_many(:actors).through(:actor_movies) }
+  end
+
+  describe "validations" do
+    it { should validate_presence_of(:title) }
+    it { should validate_presence_of(:creation_year) }
+    it { should validate_presence_of(:genre) }
+  end
```
abdulredd, line 14: Perfect validation tests!

spec/models/movie_spec.rb
```diff
@@ -2,6 +2,52 @@
...
+
+        @movie_1.actors << @actor_1
+        @movie_1.actors << @actor_2
+        @movie_1.actors << @actor_3
+
+        expect(@movie_1.actor_age_avg).to eq(31.67)
+      end
+    end
   end
 end
```
abdulredd, line 53: 👍

spec/models/studio_spec.rb
```diff
@@ -2,6 +2,53 @@
...
+        @movie_1.actors << @actor_1
+        @movie_1.actors << @actor_2
+        @movie_2.actors << @actor_1
+        @movie_2.actors << @actor_2
+        @movie_3.actors << @actor_3
+
+        expect(@studio_1.actors_all).to eq([@actor_1, @actor_2])
+        expect(@studio_2.actors_all).to eq([@actor_3])
+      end
+    end
```
abdulredd, line 52: 👍

PR #5, opened 2023-10-30
Justin Lee

JustinSteel:
Oof got caught goin down a couple wrong paths for a while.

### DIFF
--- a/app/controllers/actors_controller.rb
+++ b/app/controllers/actors_controller.rb
@@ -1,0 +1,5 @@
+class ActorsController < ApplicationController
+  def index
+    @actors = Actor.all
+  end
+end
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -1,0 +1,5 @@
+class MoviesController < ApplicationController
+  def index
+    @movies = Movie.all
+  end
+end
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -2,4 +2,8 @@ class MoviesController < ApplicationController
   def index
     @movies = Movie.all
   end
+
+  def show
+    @movie = Movie.find(params[:id])
+  end
 end
--- a/app/models/actor.rb
+++ b/app/models/actor.rb
@@ -1,0 +1,12 @@
+class Actor < ApplicationRecord
+  has_many :actor_movies
+  has_many :movies, through: :actor_movies
+
+  def self.order_by_age
+    order(age: :desc)
+  end
+
+  def average_age
+    actors.average(:age)
+  end
+end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -1,3 +1,10 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_many :actor_movies
+  has_many :actors, through: :actor_movies
 end
+
+def average_age
+  actors.average(:age)
+end
+  
--- a/spec/features/movies/show_spec.rb
+++ b/spec/features/movies/show_spec.rb
@@ -1,0 +1,29 @@
+require "rails_helper"
+
+RSpec.describe "movie show page", type: :feature do
+  it "shows movie title, creation year, and genre" do
+    studio1 = Studio.create!(name: "Universal Studios", location: "Hollywood")
+    movie1 = studio1.movies.create!(title: "Raiders of the Lost Ark", creation_year: 1981, genre: "Action/Adventure")
+    
+    visit "/movies/#{movie1.id}"
+    
+    expect(page).to have_content("Raiders of the Lost Ark")
+    expect(page).to have_content("1981")
+    expect(page).to have_content("Action/Adventure")
+  end
+
+  it "shows all actors in the movie" do
+    studio1 = Studio.create!(name: "Universal Studios", location: "Hollywood")
+    movie1 = studio1.movies.create!(title: "Raiders of the Lost Ark", creation_year: 1981, genre: "Action/Adventure")
+    actor1 = Actor.create!(name: "Harrison Ford", age: 78)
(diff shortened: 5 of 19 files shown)


app/controllers/actors_controller.rb
```diff
@@ -0,0 +1,5 @@
+class ActorsController < ApplicationController
+  def index
+    @actors = Actor.all
```
juliet-e, line 3: No need for an Actor index! The index action we're looking for is for studios (US 1). Be mindful of not creating actions that you don't need in order to not create more work for yourself!

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,9 @@
+class MoviesController < ApplicationController
+  def index
+    @movies = Movie.all
+  end
+
+  def show
+    @movie = Movie.find(params[:id])
```
juliet-e, line 7: This is the action we need! No need for the index here

app/models/actor.rb
```diff
@@ -0,0 +1,12 @@
+class Actor < ApplicationRecord
+  has_many :actor_movies
+  has_many :movies, through: :actor_movies
+
+  def self.order_by_age
+    order(age: :desc)
```
juliet-e, line 6: You're on the right track here by encapsulating this logic! We want these queries to live in a model method, but in this case we want this logic to be for a *given movie*, so we'd want to call this from a movie instance in the Movie model like so:
`self.actors.order(:age)` where self is the movie instance. 

Same with average age-- the US is looking for the avg age of actors for this given movie rather than all actors in the database

app/models/movie.rb
```diff
@@ -1,3 +1,10 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_many :actor_movies
```
juliet-e, line 3: Nice many to many set up!

app/models/movie.rb
```diff
@@ -1,3 +1,10 @@
...
+  has_many :actors, through: :actor_movies
 end
+
+def average_age
+  actors.average(:age)
```
juliet-e, line 8: Nice! This is what I meant in the previous comment! This will grab actors for a given movie rather than all actors

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,33 @@
+require "rails_helper"
+
+RSpec.describe "movie show page", type: :feature do
+  it "shows movie title, creation year, and genre" do
+    studio1 = Studio.create!(name: "Universal Studios", location: "Hollywood")
+    movie1 = studio1.movies.create!(title: "Raiders of the Lost Ark", creation_year: 1981, genre: "Action/Adventure")
+    
+    visit "/movies/#{movie1.id}"
+    
+    expect(page).to have_content("Raiders of the Lost Ark")
```
juliet-e, line 10: It's best to anchor this data to labels that describe what it refers to, and then test for the whole text. For example, expecting the page to have "Raiders of the Lost Ark, Released in 1981" is more robust than expecting just Raiders and 1981 as separate assertions. That text could be anywhere so it's more powerful to anchor these fields together. Even more robust would be to add within blocks to assert that this data appears in the section for Bros and not elsewhere

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    actor2 = Actor.create!(name: "Karen Allen", age: 69)
+    actor3 = Actor.create!(name: "Paul Freeman", age: 79)
+    ActorMovie.create!(actor: actor1, movie: movie1)
+    ActorMovie.create!(actor: actor2, movie: movie1)
+    ActorMovie.create!(actor: actor3, movie: movie1)
+
+    visit "/movies/#{movie1.id}"
+    
+    expect(page).to have_content("Karen Allen")
+    expect(page).to have_content("Harrison Ford")
```
juliet-e, line 28: Consider using orderly to be able to assert what text appears before other text

spec/features/studios/index_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      
+      # including the movie's title, creation year, and genre
+      expect(page).to have_content("Raiders of the Lost Ark")
+      expect(page).to have_content("1981")
+      expect(page).to have_content("Action/Adventure")
+      
+      
+      expect(page).to have_content("A New Hope")
+      expect(page).to have_content("1977")
+      expect(page).to have_content("Action/Adventure")
```
juliet-e, line 35: within blocks would be helpful here to verify that the movies correspond to the correct studio!

spec/models/actor_movie_spec.rb
```diff
@@ -0,0 +1,4 @@
+class ActorMovie < ApplicationRecord
+  belongs_to :actor
```
juliet-e, line 2: I think you might've copied the model here but this spec file shoud include assertions like this:
`    it {should belong_to :movie}`
`    it {should belong_to :actor}`

spec/models/actors_spec.rb
```diff
@@ -0,0 +1,22 @@
+require "rails_helper"
+
+RSpec.describe Actor, type: :model do
+  describe "relationships" do
+    it {should have_many :actor_movies}
+    it {should have_many(:movies).through(:actor_movies)}
+  end
+
+  describe "methods" do
+    it "can order actors by age" do
```
juliet-e, line 10: Nice work testing these model methods!

spec/models/movie_spec.rb
```diff
@@ -3,5 +3,7 @@
 RSpec.describe Movie, type: :model do
   describe "relationships" do
     it {should belong_to :studio}
+    it {should have_many :actors}
```
juliet-e, line 6: make sure to test your model method here!

PR #6, opened 2023-10-30
Eric Belongea

EricBelongea:
This felt easier than I was expecting, happily surprised. Gotta say, those practice ICs are suuuuper helpful. This felt on the same level as the Apollo 14 and Ticket, please practice ICs. Hoping it's a pass!

### DIFF
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -1,0 +1,5 @@
+class MoviesController < ApplicationController
+  def show  
+    @movie = Movie.find(params[:id])
+  end
+end
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -1,5 +1,10 @@
 class MoviesController < ApplicationController
   def show  
     @movie = Movie.find(params[:id])
+    if params[:actor_id]
+      @movie.add_actor(params[:actor_id])
+      
+      redirect_to "/movies/#{@movie.id}"
+    end
   end
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -10,4 +10,8 @@ class Movie < ApplicationRecord
   def average_actor_age
     actors.average(:age)
   end
+
+  def add_actor(actor_id)
+    actors << Actor.find_by(id: actor_id)
+  end
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -6,4 +6,8 @@ class Movie < ApplicationRecord
   def actors_by_age
     actors.order(:age).pluck(:name)
   end
+
+  def average_actor_age
+    actors.average(:age)
+  end
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -1,3 +1,4 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  # has_many :actors
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -1,4 +1,5 @@
 class Movie < ApplicationRecord
   belongs_to :studio
-  # has_many :actors
+  has_many :actor_movies
+  has_many :actors, through: :actor_movies
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -2,4 +2,8 @@ class Movie < ApplicationRecord
   belongs_to :studio
   has_many :actor_movies
   has_many :actors, through: :actor_movies
+
+  def actors_by_age
+    actors.order(:age).pluck(:name)
+  end
 end
--- a/app/models/studio.rb
+++ b/app/models/studio.rb
@@ -1,3 +1,7 @@
 class Studio < ApplicationRecord
   has_many :movies
+
+  def studio_movies
+    movies.pluck(:title)
+  end
 end
--- a/app/models/studio.rb
+++ b/app/models/studio.rb
@@ -6,7 +6,6 @@ class Studio < ApplicationRecord
   end
 
   def contributing_actors
-    acts = Actor.joins(:movies).where(movies: {id: movies}).distinct.pluck(:name)
-
+    Actor.joins(:movies).where(movies: {id: movies}).distinct.pluck(:name)
   end
 end
--- a/app/models/studio.rb
+++ b/app/models/studio.rb
@@ -4,4 +4,9 @@ class Studio < ApplicationRecord
   def studio_movies
     movies.pluck(:title)
   end
+
+  def contributing_actors
+    acts = Actor.joins(:movies).where(movies: {id: movies}).distinct.pluck(:name)
+
+  end
 end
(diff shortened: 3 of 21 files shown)


app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,10 @@
+class MoviesController < ApplicationController
+  def show  
+    @movie = Movie.find(params[:id])
+    if params[:actor_id]
```
cjsim89, line 4: Looks like you are adding an actor inside of a `show` method, which doesn't quite fit with SRP & RESTful actions... Can you think of a different way to separate these concerns (showing one Movie vs. adding an actor to a movie)? 

If it were me, I would make a separate controller called `movie_actors_controller` that was only responsible for actors that belonged to a movie. There, we could have a `create` action that did what you did on lines 4-7, and leave line 3 just in this `MoviesController#show`.

cjsim89: [This article](http://jeromedalbert.com/how-dhh-organizes-his-rails-controllers/) has been a helpful resource to many when figuring out how to design  Rails controllers that make sense in terms of SRP.

app/models/movie.rb
```diff
@@ -1,3 +1,17 @@
...
+
+  def actors_by_age
+    actors.order(:age).pluck(:name)
+  end
+
+  def average_actor_age
+    actors.average(:age)
+  end
+
+  def add_actor(actor_id)
```
cjsim89, line 14: This is fine but not necessary - we can do things like `<<` and `find`/`find_by` in the controller without violating MVC concerns. The other reason I would avoid this is that now you have to _test it_; but when you look at it, what are you testing? Really just the `<<` operator, and that comes standard with Ruby, so it's kinda overkill. I like the effort though!

app/models/movie.rb
```diff
@@ -1,3 +1,17 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_many :actor_movies
+  has_many :actors, through: :actor_movies
+
+  def actors_by_age
+    actors.order(:age).pluck(:name)
```
cjsim89, line 7: Excellent associations & use of AR methods!

app/models/studio.rb
```diff
@@ -1,3 +1,11 @@
 class Studio < ApplicationRecord
   has_many :movies
+
+  def studio_movies
+    movies.pluck(:title)
+  end
+
+  def contributing_actors
+    Actor.joins(:movies).where(movies: {id: movies}).distinct.pluck(:name)
```
cjsim89, line 9: Great query! 👍 

Just for fun, how would you write this query if you relied instead on AR associations...? (spoiler/hint below)


First, you'd have to add 
```ruby
has_many :actors, through: :movies
```
to this model. At which point you could do: 
```ruby
actors.distinct #.pluck(:name) optionally
```

Either approach is fine!

cjsim89: To be clear, the reason that extra `has_many` works is because this model already `has_many :movies`, so when we go to the Movie model, it already `has_many :actors`.... so, thanks to some transitive logic, a Studio would have many Actors through its Movies.

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,17 @@
...
+<h3>Actors:</h3>
+<% @movie.actors_by_age.each do |actor| %>
+  <li><%= actor %></li>
+<% end %>
+
+<h3>Average Actor Age</h3>
+<%= @movie.average_actor_age %>
+
+<h3><u>Add an Actor!</u></h3>
+<%= form_with method: :get, local: :true do |f| %>
```
cjsim89, line 13: I'm not sure I agree with the `method: :get` here, since the user story is really asking us to **create a relationship** between this movie and an actor. The associated HTTP method to a create is `post`. Using a `GET` method would work if we're only searching for an actor, but since we're creating a relationship, we should use `:post`.

app/views/studios/index.html.erb
```diff
@@ -0,0 +1,9 @@
+<h1>All Studios</h1>
+
+<% @studios.each do |studio| %>
+  <u><%= studio.name %> 
+  <%= studio.location %></u> <br>
+  <% studio.studio_movies.each do |movie| %>
```
cjsim89, line 6: You could also do `studio.movies` here since a studio already `has_many :movies`. That way, it gets straight into your movie objects and you can use them much easier.

config/routes.rb
```diff
@@ -3,4 +3,10 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/studios", to: "studios#index"
+  get "/studios/:id", to: "studios#show"
+  get "/movies/:id", to: "movies#show"
+
```
cjsim89, line 10: Using what I stated above about the `:post` method, as well as the extra controller, our route could then look like

```ruby
post "/movie_actors", to: "movie_actors#create"
```
The path could always change to `/movies/:id/actors` too, there are lots of options here. In any case, we should see a `post` method when creating.

spec/features/movie_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    @movie3 = Movie.create!(title: "Mission Impossible", creation_year: "2015", genre: "action", studio: @studio1)
+    @movie4 = Movie.create!(title: "Barbie", creation_year: "2023", genre: "comedy", studio: @studio1)
+    @movie5 = Movie.create!(title: "Oppenheimer", creation_year: "2023", genre: "documentary", studio: @studio1)
+    @movie6 = Movie.create!(title: "frozen", creation_year: "2012", genre: "animation", studio: @studio2)
+
+    @actor1 = Actor.create!(name: "Meryl Streep", age: 73)
+    @actor2 = Actor.create!(name: "Tom Hanks", age: 64)
+    @actor3 = Actor.create!(name: "Morgan Freeman", age: 82)
+    @actor4 = Actor.create!(name: "Jack Black", age: 56)
+    @actor5 = Actor.create!(name: "Margot Robbie", age: 33)
```
cjsim89, line 20: Great object setup in your tests - keep this up and use them!

spec/features/movie_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    
+    expect(@actor4.name).to appear_before(@actor3.name)
+
+    expect(page).to have_content(@movie4.average_actor_age)
+
+    expect(page).to have_content("Enter actor id")
+    expect(page).to have_button("Add Actor")
+
+    fill_in(:actor_id, with: "#{@actor5.id}")
+    expect(page).to_not have_content(@actor5.name)
```
cjsim89, line 65: Great use of `.to_not` here to test the negative case, giving more weight to the postive later on line 69.

spec/features/movie_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    click_button("Add Actor")
+
+    expect(current_path).to eq("/movies/#{@movie4.id}")
+    expect(page).to have_content(@actor5.name)
+
+    expect(@actor5.name).to appear_before(@actor4.name)
+    expect(@actor4.name).to appear_before(@actor3.name)
```
cjsim89, line 72: good use of `orderly`!

cjsim89 (commented):
Great work overall! 🙌

PR #7, opened 2023-10-30
Joseph Lee

josephlee702:
I think this went well. Finished the three user stories with 40 minutes to spare. 

Got 2 hours of sleep last night due to my ingesting coffee at 8pm....so might have missed a couple things. Looking forward to getting feedback for this but as to how I felt going through it, I felt good about it.


app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,22 @@
+class MoviesController < ApplicationController
+  def show
+    @movie = Movie.find(params[:id])
+    @actors = @movie.actors
+
+    @ordered_actors = @movie.order_actors
+    @average_age_of_cast = @movie.average_age
```
cjsim89, line 7: This is fine, but you can also access `@movie.order_actors` and any other model methods from the view - you don't have to send them all down as different instance variables from the controller. That way, you are keeping the controller action "lightweight" and easy to read and maintain.

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def update
```
cjsim89, line 10: If we are updating a movie, I would expect to see params for `title` or `age` etc.... Let's find out...

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    @movie = Movie.find(params[:id])
+
+    if params[:search].present?
+      @actor = Actor.find(params[:search])
+      @movie.add_actor(@actor)
```
cjsim89, line 15: Inside this `update` method we're adding an actor. When I look at the method you wrote called `.add_actor`, it's doing `actors << @actor`, which is the same as calling `MovieActor.create!(movie_id: ..., actor_id: ...)`. 
So, we're not really updating a movie though, are we? This means it would fit better inside a controller action called `create`; and we're not creating a movie, we're creating a `MovieActor`, so maybe it should live in a `MovieActorsController` instead. 

[This article](http://jeromedalbert.com/how-dhh-organizes-his-rails-controllers/) has been helpful when figuring out the SRP of controllers and different approaches we can take in Rails.

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+      @ordered_actors = @movie.order_actors
+      @average_age_of_cast = @movie.average_age
```
cjsim89, line 17: These two instance variables are unused since we are redirecting to the show page, which just loads them again.

app/models/movie.rb
```diff
@@ -1,3 +1,33 @@
...
+  def order_actors
+    actors.order(:age)
+  end
+
+  def average_age
+    ages = []
+    actors.each do |actor|
+      ages << actor.age
+    end
+    average = ages.sum / ages.length.to_f.round(1)
```
cjsim89, line 23: This approach is using Ruby instead of AR. The main red flag for me is when I see `.each` inside a model method. 

If you're looking to find an average of a given column, try using `.average` instead. Ex: 

```ruby
actors.average(:age)
```
This method will look at the `actors` association present in this model, and perform an average calculation of the given attribute on that model. An Actor instance has an `age` property of numeric value, so it can average them together for us.

app/models/movie.rb
```diff
@@ -1,3 +1,33 @@
...
+    average
+  end
+
+  def add_actor(actor)
+    if !actors.include?(actor)
```
cjsim89, line 28: Does the user story specify that only one instance of an actor is allowed to be in this movie? What about "The Wizard of Oz" or "Austin Powers"? Those actors may be one person but they had multiple roles... 🤔 food for thought. 

In any case, this line is not needed. It's not currently up to us (given the user stories) if we need this check, so don't include it until we have to.

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,19 @@
+<h2><u><%= @movie.concat_movie_details %></u></h2>
+
+<%= form_with url: "/movies/#{@movie.id}", method: :patch, local: true do |f| %>
```
cjsim89, line 3: See above - I would rather use `:post` here than `:patch` since we're not updating a movie, we're creating an instance of MovieActor.

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    @new_actor = Actor.create!(name: "Joseph Lee", age: 26)
+  end
+
+  it "shows the employee's name, department, and list of tickets from oldest to newest" do
+    visit "/movies/#{@mononoke.id}"
+    expect(page).to have_content("#{@mononoke.title} - #{@mononoke.creation_year} - #{@mononoke.genre}")
+    expect(page).to have_content("Cast")
+    expect(page).to have_content(@mononoke_actor.name)
+    expect(page).to have_content(@mononoke_actress.name)
+    expect(@mononoke_actress.name).to be < @mononoke_actor.name
```
cjsim89, line 18: I'm.. not sure what this `<` operator is doing..? Is it really "less than" or do you mean, it comes before the actor's name on the page...?

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  it "adds the searched actor/actress to the movie and calculates average age accordingly" do
+    visit "/movies/#{@mononoke.id}"
+
+    fill_in "Search", with: "#{@new_actor.id}"
+    click_on("Search")
+
+    expect(current_path).to eq("/movies/#{@mononoke.id}")
+    expect(page).to have_content(@new_actor.name)
+    expect(page).to have_content(@new_actor.age)
+    #these lines under just indicate that "Joseph Lee" should appear before the other two cast members' names
```
cjsim89, line 36: Understood - I recommend using the [`orderly` gem](https://github.com/jmondo/orderly) - it is way way better syntax for finding order of items on a page in Capybara.

spec/models/movie_spec.rb
```diff
@@ -1,7 +1,56 @@
...
+      expect(@mononoke.actors).to eq([@mononoke_actor, @mononoke_actress])
+      expect(@mononoke_actor.age).to eq(45)
+      expect(@mononoke_actress.age).to eq(37)
+      expect(@mononoke.average_age).to eq(41.0)
+
+      @mononoke.add_actor(@new_actor)
+      expect(@mononoke.actors).to eq([@mononoke_actor, @mononoke_actress, @new_actor])
+      expect(@mononoke.average_age).to eq(36.0)
+
+      @mononoke.add_actor(@new_actress)
```
cjsim89, line 51: Good unit & feature tests overall!

cjsim89 (commented):
Good work overall! I think this is pretty close; the main issue is using Ruby instead of AR in that one area, and one MVC concern. Try rewriting your Ruby into AR for good practice!

spec/features/movies/show_spec.rb
josephlee702: comes before! I found it in an article a while ago and had been writing tests like that. Would I be dinged for that?

spec/features/movies/show_spec.rb
josephlee702: oh - just saw this comment. got it.

PR #8, opened 2023-10-30
Blake S

bserge13:
Went very well overall. Only a few things needed reaching-out for, but glad I did for future takeaways. Thank you both!

### DIFF
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -1,0 +1,5 @@
+class MoviesController < ApplicationController
+  def show
+    @movie = Movie.find(params[:id]) 
+  end
+end
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -1,5 +1,13 @@
 class MoviesController < ApplicationController
   def show
     @movie = Movie.find(params[:id]) 
+    @actors = @movie.actor_order_age
+
+    if
+      params[:name].present?
+      new_actor = Actor.find_by(name: params[:name])
+      @movie.actors << new_actor 
+      redirect_to "/movies/#{@movie.id}"
+    end 
   end
 end
--- a/app/models/actor.rb
+++ b/app/models/actor.rb
@@ -1,0 +1,4 @@
+class Actor < ApplicationRecord
+  has_many :movie_actors
+  has_many :movies, through: :movie_actors
+end
--- a/app/models/actor.rb
+++ b/app/models/actor.rb
@@ -4,4 +4,8 @@ class Actor < ApplicationRecord
 
   validates :name, presence: true 
   validates :age, presence: true, numericality: true 
+
+  def self.average_age
+    average(:age).round(2)
+  end
 end
--- a/app/models/actor.rb
+++ b/app/models/actor.rb
@@ -1,4 +1,7 @@
 class Actor < ApplicationRecord
   has_many :movie_actors
   has_many :movies, through: :movie_actors
+
+  validates :name, presence: true 
+  validates :age, presence: true, numericality: true 
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -1,3 +1,5 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_many :movie_actors
+  has_many :actors, through: :movie_actors
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -6,4 +6,8 @@ class Movie < ApplicationRecord
   validates :title, presence: true 
   validates :creation_year, presence: true 
   validates :genre, presence: true 
+
+  def actor_order_age
+    self.actors.order(age: :asc) 
+  end
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -2,4 +2,8 @@ class Movie < ApplicationRecord
   belongs_to :studio
   has_many :movie_actors
   has_many :actors, through: :movie_actors
+
+  validates :title, presence: true 
+  validates :creation_year, presence: true 
+  validates :genre, presence: true 
 end
--- a/app/views/movies/show.html.erb
+++ b/app/views/movies/show.html.erb
@@ -3,3 +3,19 @@
 <%= @movie.title %>
 <%= @movie.creation_year %>
 <%= @movie.genre %>
+
+<p> Actors: </p>
+
+<p><% @actors.each do |actor| %>
+  <li>Name: <%= actor.name %>
+  Age: <%= actor.age %></li>
+  <% end %></p>
+  <p> Actors Average Age: <%= @actors.average_age %></p>
+
+
+  <h3> Add a Actor to this Movie: </h3>
+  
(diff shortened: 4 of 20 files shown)


app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,13 @@
+class MoviesController < ApplicationController
+  def show
+    @movie = Movie.find(params[:id]) 
+    @actors = @movie.actor_order_age
```
juliet-e, line 4: Just FYI, this second instance var isn't strictly necessary as it's totally okay to call `@movie.actor_order_age` in the view itself! That's not an MVC violation. An MVC violation would be calling Active Record query methods in the view, but it's okay to call a model method name

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+
+    if
+      params[:name].present?
```
juliet-e, line 7: Rather than routing the form back to this action, a more RESTful option that better adheres to SRP would be to route it to a `create` action. In this case we are creating an association between an existing movie and an existing actor. So, we could create a joins controller called MovieActorsController and route the form to a `create` action there. That action would hold this logic - finding the existing movie and actor and then adding the actor to that movie. Then, it can redirect back to the movie show page. 

Without breaking up this controller action, this action is doing too much - it handles too much responsibility (SRP violation)

app/models/actor.rb
```diff
@@ -0,0 +1,11 @@
+class Actor < ApplicationRecord
+  has_many :movie_actors
+  has_many :movies, through: :movie_actors
+
+  validates :name, presence: true 
+  validates :age, presence: true, numericality: true 
+
+  def self.average_age
+    average(:age).round(2)
```
juliet-e, line 9: This US was actually asking to see all the given *movie instance's* actors rather than all actors in the database. So, a better spot for this would be in the Movie model where we could call `actors.average(:age).round(2)`

app/models/movie.rb
```diff
@@ -1,3 +1,13 @@
...
   belongs_to :studio
+  has_many :movie_actors
+  has_many :actors, through: :movie_actors
+
+  validates :title, presence: true 
+  validates :creation_year, presence: true 
+  validates :genre, presence: true 
+
+  def actor_order_age
+    self.actors.order(age: :asc)
```
juliet-e, line 11: Nice work encapsulating this data logic!

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,21 @@
...
+<p><% @actors.each do |actor| %>
+  <li>Name: <%= actor.name %>
+  Age: <%= actor.age %></li>
+  <% end %></p>
+  <p> Actors Average Age: <%= @actors.average_age %></p>
+
+
+  <h3> Add a Actor to this Movie: </h3>
+  
+  <%= form_with url: "/movies/#{@movie.id}?actor_name=#{params[:name]}&commit=Search", method: :get do %>
```
juliet-e, line 18: You don't need to use query parameters here if you use a route that can carry the movie ID, and then keep in mind that the form field for the actor (which should be ID, not name!) will have the actor info so that doesn't need to be in this url either. 
`form_with url: "/movies/:movie_id/actors", method: :post, data: {turbo: false} do |form|`

This is something I remember from your pair project as well, so I encourage you to refactor it in this IC or in your project (or both!) and add a note about this in a cheat sheet so that you don't continue to format forms like this!

app/views/studios/index.html.erb
```diff
@@ -0,0 +1,13 @@
+<h2> Studios Index Page </h2>
+
+  <% @studios.each do |studio| %>
+    <section id=<%= "studio-#{studio.id}" %>>
```
juliet-e, line 4: Nice section usage!

config/routes.rb
```diff
@@ -3,4 +3,11 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  # studios 
+  get "/studios", to: "studios#index" 
+
+  # movies 
+  get "/movies/:id", to: "movies#show"
+
```
juliet-e, line 12: Given the form route refactor, you would need to add a POST route that directs to a joins controller #create action:
`post "/movies/:id/actors", to: "movie_actors#create"`

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    expect(page).to have_content(@allen.name)
+    expect(page).to have_content(@allen.age)
+    expect(page).to have_content(@reeves.name)
+    expect(page).to have_content(@reeves.age)
+
+    expect(@reeves.name).to appear_before(@allen.name)
+    expect(@allen.name).to appear_before(@hanks.name)
+
+    expect(page).to have_content("Actors Average Age:")
+    expect(Actor.average_age).to eq(41.0)
```
juliet-e, line 36: When we're testing this functionality, we don't want to call the production code that we're testing. In this case, we're testing the `average_age` method, so it'd be better to expect something like
`expect(page).to have_content("Actors Average Age: 41")` rather than separating these two assertions and calling the model method in the assertion

juliet-e: We want to do the calculation and know what to expect, and then by asserting this expected value (41), we are verifying that the production code under the hood is running this calculation correctly

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    pratt = Actor.create!(name: "Chris Pratt", age: 35)
+
+    visit "/movies/#{@toy_1.id}"
+
+    fill_in "Enter actor name", with: "Chris Pratt"
+    click_button "Add Actor" 
+
+    expect(current_path).to eq("/movies/#{@toy_1.id}")
+    expect(page).to have_content("Chris Pratt")
+    expect(Actor.average_age).to eq(39.5)
```
juliet-e, line 57: same here! We want to expect the display of "Actors Average Age: 39.5" rather than calling the production code here!

spec/features/studios/index_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      expect(page).to have_content(@toy_1.title)
+      expect(page).to have_content(@toy_1.creation_year)
+      expect(page).to have_content(@toy_1.genre)
+      
+      expect(page).to have_content(@toy_2.title)
+      expect(page).to have_content(@toy_2.creation_year)
+      expect(page).to have_content(@toy_2.genre)
+    end 
+
+    within "#studio-#{@mgm.id}" do
```
juliet-e, line 33: Nice use of within blocks!

spec/models/movie_spec.rb
```diff
@@ -2,6 +2,32 @@
...
+        allen = Actor.create!(name: "Tim Allen", age: 41)
+        reeves = Actor.create!(name: "Keanu Reeves", age: 40)
+    
+        MovieActor.create!(movie_id: toy_1.id, actor_id: hanks.id)
+        MovieActor.create!(movie_id: toy_1.id, actor_id: allen.id)
+        MovieActor.create!(movie_id: toy_1.id, actor_id: reeves.id)
+    
+        expect(toy_1.actor_order_age).to eq([reeves, allen, hanks])
+      end
+    end
```
juliet-e, line 31: Nice work with model testing!

PR #9, opened 2023-10-30
Sooyung Kim

skim1027:
It was really hard... got really confused on US3, tried looking back at the last week's feedback on the create.. was confused for a while but I think I figured it out?  Didn't have enough time to look at the extension.


app/controllers/movie_actors_controller.rb
```diff
@@ -0,0 +1,8 @@
+class MovieActorsController < ApplicationController
+  def create
+    movie = Movie.find(params[:id])
+    new_actor = Actor.find(params[:actor_id])
+    MovieActor.create!(movie: movie, actor: new_actor)
+    redirect_to "/movies/#{movie.id}"
+  end
+end
```
abdulredd, line 8: You got it! The only thing I would mention is that this isn't necessarily a "new actor" so the naming could be confusing, but the implementation is correct. Great job.

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,5 @@
+class MoviesController < ApplicationController
+  def show
+    @movie = Movie.find(params[:id])
+  end
+end
```
abdulredd, line 5: 👍

app/controllers/studios_controller.rb
```diff
@@ -0,0 +1,5 @@
+class StudiosController < ApplicationController
+  def index
+    @studios = Studio.all
+  end
+end
```

app/models/movie.rb
```diff
@@ -1,3 +1,13 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_many :movie_actors
+  has_many :actors, through: :movie_actors
+
+  def actors_list
+    actors.order(age: :asc)
+  end
```
abdulredd, line 8: Great job with this, just note that ascending is the default, so `actors.order(:age)` is the same thing. :)

app/models/movie.rb
```diff
@@ -1,3 +1,13 @@
...
+
+  def ave_actors_age
+    actors.average(:age).ceil
+  end
```
abdulredd, line 12: I just want to make this quick note here that depending on who you ask, doing something like a `ceil` here could be considered presentation logic. I won't mark this as an MVC infraction, but something I wanted to bring up because some rubyist would consider this an infraction because it is technically presentation logic, I am in that camp.

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,21 @@
...
+<p>Average Actors Age: <%= @movie.ave_actors_age %></p>
+
+
+<%= form_with url: "/movies/#{@movie.id}/actors", method: :post, data: {turbo:false},local: true do |form| %>
+  
+  <%= form.label :actor_id, "Actor ID" %>
+  <%= form.text_field :actor_id %>
+
+  <%= form.submit "Submit" %>
+<% end %>
```
abdulredd, line 21: Yes! This form is perfect and follows restful conventions perfectly! You're using the correct verb (POST) and sending the data to the correct path, great job Sooyung!

config/routes.rb
```diff
@@ -3,4 +3,10 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/studios", to: "studios#index"
+
+  get "/movies/:id", to: "movies#show"
+
+  post "/movies/:id/actors", to: "movie_actors#create"
 end
```

app/views/studios/index.html.erb
```diff
@@ -0,0 +1,14 @@
...
+<p>Studio Name: <%= studio.name %></p>
+<p>Studio Location: <%= studio.location %></p></div>
+
+  <div id="movie-<%= studio.id %>">
+  <%studio.movies.each do |movie| %>
+    <p>Movie Title: <%= movie.title %></p>
+    <p>Movie Creation Year: <%= movie.creation_year %></p>
+    <p>Movie Genre: <%= movie.genre %></p>
+  <% end %>
+<% end %>
```
abdulredd, line 14: 👍

db/migrate/20231030151038_create_movie_actors.rb
```diff
@@ -0,0 +1,10 @@
+class CreateMovieActors < ActiveRecord::Migration[7.0]
+  def change
+    create_table :movie_actors do |t|
+      t.references :movie, null: false, foreign_key: true
+      t.references :actor, null: false, foreign_key: true
+
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 10: Great job setting up your join table!

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+        fill_in("Actor ID", with: @john.id)
+
+        click_button("Submit")
+
+        expect(current_path).to eq("/movies/#{@marry.id}")
+        expect(page).to have_content(@john.name)
+      end
+    end
+  end
+end
```
abdulredd, line 58: Wow, I really like these tests.... they remove any ambiguity and say explicitly what is being tested. For example, I know just by looking at your tests that 44 is supposed to show up on the page as the average age of cast members. Keep this up, your future teammates will love you for this.

spec/features/studios/index_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+          expect(page).to have_content('Movie Creation Year: 2000')
+          expect(page).to have_content('Movie Genre: Action')
+          expect(page).to have_content('Movie Title: Kung Fu Panda')
+          expect(page).to have_content('Movie Creation Year: 2008')
+          expect(page).to have_content('Movie Genre: Anime')
+        end
+      end
+    end
+  end
+end
```
abdulredd, line 54: Your tests are amazing, great use of the `within` block to further reduce ambiguity in what is being tested and where.

spec/models/movie_spec.rb
```diff
@@ -1,7 +1,37 @@
...
+    it 'lists the actors from youngest to oldest' do
+      expect(@marry.actors_list).to eq([@chloe, @sarah, @jenni, @owen])
+    end
+  end
+
+  describe 'ave_actors_age' do
+    it 'averages actors age for the movie' do
+      expect(@marry.ave_actors_age).to eq(44)
+    end
   end
```
abdulredd, line 36: Great job testing your instance methods. Like I mentioned before I personally wouldn't have used `ceil` in my logic inside of my method and instead would have used it in my view and in my tests. My tests would have looked something like this:

```ruby
expect(@marry.ave_actors_age.ceil).to eq(44)
```

abdulredd:
This is a perfect submission Sooyung, if you hadn't mentioned in your reflection that this was hard for you I would have never guessed. Keep up the great work!

PR #10, opened 2023-10-30
Brendan Bondurant

brendan-bondurant:
I got through the extensions, and I think it went pretty well.

### DIFF
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -1,0 +1,13 @@
+class MoviesController < ApplicationController
+
+  def show
+    @movie = Movie.find(params[:id])
+  end
+
+  def update
+    movie = Movie.find(params[:id])
+    actor = Actor.find(params[:add_actor])
+    movie.add_actor(actor)
+    redirect_to "/movies/#{movie.id}"
+  end
+end
--- a/app/models/actor.rb
+++ b/app/models/actor.rb
@@ -1,0 +1,4 @@
+class Actor < ApplicationRecord
+  has_many :movie_actors
+  has_many :movies, through: :movie_actors
+end
--- a/app/models/actor.rb
+++ b/app/models/actor.rb
@@ -1,4 +1,8 @@
 class Actor < ApplicationRecord
   has_many :movie_actors
   has_many :movies, through: :movie_actors
+
+  def self.average_age
+    self.average(:age).round
+  end
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -1,3 +1,5 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_many :movie_actors
+  has_many :actors, through: :movie_actors
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -2,4 +2,13 @@ class Movie < ApplicationRecord
   belongs_to :studio
   has_many :movie_actors
   has_many :actors, through: :movie_actors
+
+  def staff
+    self.actors.order(age: :desc)
+  end
+
+  def add_actor(actor)
+    self.actors << actor
+  end
+
 end
--- a/app/models/studio.rb
+++ b/app/models/studio.rb
@@ -3,6 +3,6 @@ class Studio < ApplicationRecord
   has_many :actors, through: :movies
 
   def actor_list
-    actors.distinct.pluck(:name)
+    actors.distinct.pluck(:name).join(", ")
   end
 end
--- a/app/models/studio.rb
+++ b/app/models/studio.rb
@@ -1,3 +1,8 @@
 class Studio < ApplicationRecord
   has_many :movies
+  has_many :actors, through: :movies
+
+  def actor_list
+    actors.distinct.pluck(:name)
+  end
 end
--- a/app/views/movies/show.html.erb
+++ b/app/views/movies/show.html.erb
@@ -1,0 +1,16 @@
+
+<h2> <%= @movie.title %> </h2>
+<p>Released in: <%=@movie.creation_year%></p>
+<p>Genre: <%=@movie.genre  %></p>
+<h4>Actors</h4>
+<p><% @movie.staff.each do |actor| %></p>
+    <p>Name:<%= actor.name %> </p>
+    <p>Age:<%= actor.age %></p>
+    <% end %>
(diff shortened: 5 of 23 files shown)


app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,13 @@
+class MoviesController < ApplicationController
+
+  def show
+    @movie = Movie.find(params[:id])
+  end
+
+  def update
```
juliet-e, line 7: This implementation looks good but a more RESTful routing option here would be to route this form to a joins contoller like `MovieActorsController #create` to indicate that you are creating an association between an existing movie and an actor. We want to reserve the movies#update action for updating attributes on a movie resource itself.

app/models/actor.rb
```diff
@@ -0,0 +1,8 @@
+class Actor < ApplicationRecord
+  has_many :movie_actors
+  has_many :movies, through: :movie_actors
+
+  def self.average_age
+    self.average(:age).round
```
juliet-e, line 6: well done encapsulating this logic in a model method! However, this is going to get the average age of *all* actors rather than just the actors in a given movie. We were looking for this to be an instance method on the movie model to average just the actors associated with that movie (or you could pass in a movie_id here)

app/models/movie.rb
```diff
@@ -1,3 +1,14 @@
...
   belongs_to :studio
+  has_many :movie_actors
+  has_many :actors, through: :movie_actors
+
+  def staff
+    self.actors.order(age: :desc)
+  end
+
+  def add_actor(actor)
+    self.actors << actor
```
juliet-e, line 11: This looks good, but it is not strictly necessary to encapsulate this logic in a model method. This *could* live in the controller but is not wrong to put here

app/models/movie.rb
```diff
@@ -1,3 +1,14 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_many :movie_actors
```
juliet-e, line 3: Great M2M relationship set up!

app/models/studio.rb
```diff
@@ -1,3 +1,8 @@
 class Studio < ApplicationRecord
   has_many :movies
+  has_many :actors, through: :movies
```
juliet-e, line 3: Nice job with the extension!

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,16 @@
...
+<p>Released in: <%=@movie.creation_year%></p>
+<p>Genre: <%=@movie.genre  %></p>
+<h4>Actors</h4>
+<p><% @movie.staff.each do |actor| %></p>
+    <p>Name:<%= actor.name %> </p>
+    <p>Age:<%= actor.age %></p>
+    <% end %>
+<p>"Average Age of Actors: <%=@movie.actors.average_age%></p>
+
+<%= form_with url: "/movies/#{@movie.id}", method: :post, local: true, data: { turbo: false } do |form| %>
```
juliet-e, line 12: A more RESTful path here could be: POST "/movies/:id/actors" to indicate creating an association between these 2 resources

config/routes.rb
```diff
@@ -1,6 +1,11 @@
 Rails.application.routes.draw do
   # Define your application routes per the DSL in https://guides.rubyonrails.org/routing.html
+  get '/studios', to: 'studios#index'
 
+  get '/movies/:id', to: 'movies#show'
+  post '/movies/:id', to: 'movies#update'
```
juliet-e, line 6: see notes above about a more RESTful path and controller action

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  it 'has a form to add an actor' do
+    visit "movies/#{@turning_red.id}"
+    expect(page).to_not have_content(@actor1.name)
+    fill_in "Add actor", with: "#{@actor1.id}"
+    click_button "Submit"
+    expect(current_path).to eq("/movies/#{@turning_red.id}")
+    expect(page).to have_content(@actor1.name)
+  end
+end
```
juliet-e, line 77: Great, robust testing!

spec/features/studios/index_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    expect(page).to have_content(@universal.name)
+    expect(page).to have_content(@disney.location)
+    expect(page).to have_content(@pixar.location)
+    expect(page).to have_content(@universal.location)
+  end
+  
+  it 'displays each movie associated with the studio' do
+    visit "/studios"
+    
+    within("section#studio-#{@disney.id}") do
```
juliet-e, line 35: Nice use of within blocks to make these assertions more meaningful!

spec/models/actor_spec.rb
```diff
@@ -0,0 +1,8 @@
+require "rails_helper"
+
+RSpec.describe Actor, type: :model do
+  describe "relationships" do
+    it { should have_many :movie_actors}
+    it { should have_many(:movies).through(:movie_actors) }
+  end
+end
```
juliet-e, line 8: Your association tests look good, but we'd love to see model tests for each of your instance or class methods!

app/models/studio.rb
```diff
@@ -1,3 +1,8 @@
...
+
+  def actor_list
+    actors.distinct.pluck(:name).join(", ")
```
juliet-e, line 6: Adding `.join` here is a gray area MVC-wise -- we can argue that it's valuable to put in the model if we want to granularly test it at a unit level, but folks may also say that it is more display logic and should live in the view. 

As long as you unit test it, my tendency is to err on the side of encapsulating it in a model method, but wanted to let you know that it's a gray area! Just make sure you do test it :)

app/models/actor.rb
brendan-bondurant: When I tested this it was for just the shrek movie, and the average age of the actors for that is 30, but the average age for all of the actors was 40...am I missing something? I added another test with a different movie and both tests passed.

juliet-e: Ahh that's because you're calling the method on the collection of actors in the view: `@movie.actors.average_age` so this is actually fine if you call it on the movie instance because it will be bound to just actors in that movie. Either way works. I missed that this was how you were using the method in the view - my bad!

juliet-e: If you had done `Actor.average_age` in the view, you would have had an incorrect answer. Still be sure to model test this method though!

PR #12, opened 2023-10-30
Scott DeVoss

scottdevoss:
I thought this went better than the last IC. I was able to make the appropriate changes to the mistakes I did last IC. I was close to finishing user story 3 and felt pretty good with my progress.

### DIFF
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -1,3 +1,5 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_many :movie_actors
+  has_many :actors, through: :movie_actors
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -6,4 +6,8 @@ class Movie < ApplicationRecord
   def sort_actors_by_age
     actors.order(age: :asc)
   end
+
+  def average_age
+    actors.average(:age).to_i
+  end
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -2,4 +2,8 @@ class Movie < ApplicationRecord
   belongs_to :studio
   has_many :movie_actors
   has_many :actors, through: :movie_actors
+
+  def sort_actors_by_age
+    actors.order(age: :asc)
+  end
 end
--- a/app/views/movies/show.html.erb
+++ b/app/views/movies/show.html.erb
@@ -1,0 +1,5 @@
+<h1>Title: <%= @movie.title %></h1><br>
+<h2>Creation Year: <%= @movie.creation_year %><br>
+Genre: <%= @movie.genre %></h2>
+
+<h3>All Actors:</h3>
--- a/app/views/movies/show.html.erb
+++ b/app/views/movies/show.html.erb
@@ -8,3 +8,9 @@ Genre: <%= @movie.genre %></h2>
 <% end %>
 
 <h3>Average Age: <%= @movie.average_age %>
+
+<%= form_with url: "/movies/#{@movie.id}", method: :patch, data: { turbo: false } do |form| %>
+  <%= form.label "Add an actor to this movie" %>
+  <%= form.text_field :actor_id %>
+  <%= form.submit "Submit" %>
+<% end %>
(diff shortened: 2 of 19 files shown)


app/models/movie.rb
```diff
@@ -1,3 +1,13 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_many :movie_actors
+  has_many :actors, through: :movie_actors
+
+  def sort_actors_by_age
+    actors.order(age: :asc)
```
cjsim89, line 7: Good AR and associations in your models!

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,16 @@
...
+Genre: <%= @movie.genre %></h2>
+
+<h3>All Actors:</h3>
+<% @movie.sort_actors_by_age.each do |actor| %>
+  <%= actor.name %><br>
+<% end %>
+
+<h3>Average Age: <%= @movie.average_age %>
+
+<%= form_with url: "/movies/#{@movie.id}", method: :patch, data: { turbo: false } do |form| %>
```
cjsim89, line 12: Good start to your form, though I'm not sure I agree with the `method: :patch`. `PATCH` (or `PUT`) is used when we want to update a resource, while `POST` will be creating a resource. The user story says we're going to add an actor to this movie, so are we really updating the `movie` resource? 
If so, then that means we should change one of the values in the `movies` table... but that's not where the actors live. If we're going to add an actor to a movie, we'd be _creating_ a new row in the join table. So, really, this becomes a `:post` method instead. 


This decision will also dictate what your routes & controllers look like, so I recommend coming back to finish this one out! But your form setup otherwise looks great 👍

spec/features/movie/show_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+        @jeff_goldblum = Actor.create!(name: "Jeff Goldblum", age: 71)
+        @jim_carrey = Actor.create!(name: "Jim Carrey", age: 61)
+        @christian_bale = Actor.create!(name: "Christian Bale", age: 49)
+        @morgan_freeman = Actor.create!(name: "Morgan Freeman", age: 86)
+        @cillian_murphy = Actor.create!(name: "Cillian Murphy", age: 47)
+
+        MovieActor.create!(movie_id: @the_dark_knight.id, actor_id: @christian_bale.id)
+        MovieActor.create!(movie_id: @the_dark_knight.id, actor_id: @morgan_freeman.id)
+        MovieActor.create!(movie_id: @the_dark_knight.id, actor_id: @cillian_murphy.id)
```
cjsim89, line 22: Great test setup!

spec/features/movie/show_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+          expect(page).to have_content(@the_dark_knight.creation_year)
+          expect(page).to have_content(@the_dark_knight.genre)
+
+          expect(page).to_not have_content(@bruce_almighty.title)
+          expect(page).to_not have_content(@jurassic_park.title)
+
+          expect(page).to have_content("All Actors")
+
+          expect(@cillian_murphy.name).to appear_before(@christian_bale.name)
+          expect(@christian_bale.name).to appear_before(@morgan_freeman.name)
```
cjsim89, line 40: Really nice feature tests and use of `orderly`

spec/features/movie/show_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      end
+
+      #User Story 3
+      describe "I do not see any actors listed that are not part of the movie and I see a form to add an actor to this movie" do
+        describe "When I fill in the form with an ID of an actor and click submit" do
+          it "I am redirected back to the movie's show page and I see the new actors name lised" do
+
+            visit "/movies/#{@the_dark_knight.id}"
+
+            expect(page).to_not have_content(@jim_carrey.name)
```
cjsim89, line 53: Great job using `.to_not` to better prove your case on lines 61-62

cjsim89 (commented):
Great work overall! Keep up your awesome test writing, and try to practice the form part of user story 3 more just to get it under your fingers. If you haven't tried them already, look at the Practice Assessments on the mod 2 website to get more practice making those form/method decisions.

PR #13, opened 2023-10-30
Kevin Zolman

zkevkev:
I think this went pretty well. I did use some ruby for the first story and had some issues with one of my paths, but aside from that this was a good opportunity to stretch my legs a bit with some basic AR.

### DIFF
--- a/app/models/studio.rb
+++ b/app/models/studio.rb
@@ -10,6 +10,6 @@ class Studio < ApplicationRecord
   end
 
   def list_actors
-    movies.joins(:actors).pluck(:name).to_sentence
+    movies.joins(:actors).distinct.pluck(:name).to_sentence
   end
 end
--- a/app/models/studio.rb
+++ b/app/models/studio.rb
@@ -1,3 +1,11 @@
 class Studio < ApplicationRecord
   has_many :movies
+
+  def list_movies_details
+    movies_details = movies.pluck(:title, :creation_year, :genre)
+
+    movies_details = movies_details.flat_map do |movie|
+      movie.join(", ")
+    end
+  end
 end
--- a/app/models/studio.rb
+++ b/app/models/studio.rb
@@ -8,4 +8,8 @@ class Studio < ApplicationRecord
       movie.join(", ")
     end
   end
+
+  def list_actors
+    
+  end
 end
--- a/app/models/studio.rb
+++ b/app/models/studio.rb
@@ -10,6 +10,6 @@ class Studio < ApplicationRecord
   end
 
   def list_actors
-    
+    movies.joins(:actors).pluck(:name).to_sentence
   end
 end
--- a/app/views/movies/show.html.erb
+++ b/app/views/movies/show.html.erb
@@ -5,8 +5,9 @@
 Average Actor Age: <%= @movie.ave_actor_age %> years
 
 Add an Actor
-<%= form_with url: "/movies/#{@movie.id}/actor/#{params[:actor_id]}", method: :post, local: true do |form| %>
+<%= form_with method: :post, local: true do |form| %>
   <%= form.label :actor_id %>
   <%= form.number_field :actor_id %>
+  <%= form.hidden_field(:value => params[:actor_id]) %>
   <%= form.submit "submit" %>
 <% end %>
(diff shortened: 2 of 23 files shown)


app/models/studio.rb
```diff
@@ -1,3 +1,15 @@
 class Studio < ApplicationRecord
   has_many :movies
+
+  def list_movies_details
+    movies_details = movies.pluck(:title, :creation_year, :genre)
+
+    movies_details = movies_details.flat_map do |movie|
+      movie.join(", ")
```
cjsim89, line 8: This is more of a "formatting" type method, which is not necessarily the purview of the Model - maybe a helper method in a module, or even a private method in a model or controller. But as such, there's nothing really testable about this method, so while it is cool, it's a bit overkill for a model since it's more about view formatting than retrieving a specific collection from your database.

app/models/studio.rb
```diff
@@ -1,3 +1,15 @@
...
+    end
+  end
+
+  def list_actors
+    movies.joins(:actors).distinct.pluck(:name).to_sentence
```
cjsim89, line 13: Nice AR - and I like the use of `to_sentence`

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,13 @@
...
+<%= @movie.creation_year %></br>
+<%= @movie.genre %></br>
+<%= @movie.list_actors %></br>
+Average Actor Age: <%= @movie.ave_actor_age %> years</br>
+
+Add an Actor
+<%= form_with method: :post, local: true do |form| %>
+  <%= form.label :actor_id %>
+  <%= form.number_field :actor_id %>
+  <%= form.hidden_field(:value => params[:actor_id]) %>
```
cjsim89, line 11: Look closely - you are asking the user for the `actor_id` in line 10, but you are providing this hidden_field with the value of a `params[:actor_id]` which is not correct. You should instead be providing it with `@movie.id`. As such, this is likely only working because your test data is limited in its creation of movies-to-actors.

cjsim89: You also forgot the `url:` in your `form_with` field; luckily it will default to the same location, but you could have avoided this issue if you included `url: "/movies/#{@movie.id}"`.

config/routes.rb
```diff
@@ -1,6 +1,10 @@
...
+  get "/studios", to: "studios#index"
+  get "/studios/:id", to: "studios#show"
 
-  # Defines the root path route ("/")
-  # root "articles#index"
+  get "/movies/:id", to: "movies#show"
+
+  # I wanted to do the path /movies/:movie_id/actors/:actor_id but wasn't
+  # able to get it to work, not sure why. Guidance here would be appreciated.
+  post "/movies/:id", to: "actor_movies#create"
```
cjsim89, line 9: The path you were trying to get to work looks good; this one also looks good. Do you remember what kind of error you were getting when it wasn't working? My guess is it had something to do with the `params[:actor_id]` in your form, but hard to tell from here.

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    visit "/movies/#{@movie_1.id}"
+
+    expect(@actor_2.name).to appear_before(@actor_1.name)
+    expect(page).to have_content("Average Actor Age: 68.5 years")
+  end
+
+  it "allows an actor to be added to the movie by submitting the actor's id" do
+    visit "/movies/#{@movie_1.id}"
+
+    expect(page).to_not have_content(@actor_3.name)
```
cjsim89, line 36: Good use of `.to_not`

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    expect(page).to have_content(@movie_1.creation_year)
+    expect(page).to have_content(@movie_1.genre)
+    expect(page).to have_content(@actor_1.name)
+    expect(page).to have_content(@actor_2.name)
+  end
+
+  it "displays actors ordered from youngest to oldest and average actor age" do
+    visit "/movies/#{@movie_1.id}"
+
+    expect(@actor_2.name).to appear_before(@actor_1.name)
```
cjsim89, line 29: Good use of `orderly`

cjsim89 (commented):
Good strong showing overall! The one area to improve on would be watching what params you send in your forms, and making sure all of your form details are there. I would call this a pretty close one overall, so you are on the right track but keep working at it!

PR #15, opened 2023-10-30
Nathan Trautenberg

WagglyDessert:
Hi Juliet, a couple comments/concerns:

1) For some reason, after the many-to-many migration, my studio index page no longer displays the movie information on the local server, but the tests still pass

2) I nearly finished user story 3, but it couldn't find an id for the actor. I couldn't figure that out, but if I could, I would've been done with the story.

3) I can't got to a movies show page on my local server. I don't know what any movie ids are, and random numbers I tried didn't get anything to populate so I wrote everything on my tests and code without being able to check if it was working as intended via a server. That was kind of frustrating. How do I fix this?

### DIFF
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -4,6 +4,17 @@ class MoviesController < ApplicationController
     @movie = Movie.find(params[:id])
     @actor_names = @movie.order_by_age
     @avg_age = @movie.avg_age
+    if params[:search_by_name].present? 
+      threshold = params[:search_by_name].to_s
+      @actors = Actor.filter_by_name(threshold)
+    end
+  end
+
+  def update
+    @movie = Movie.find(params[:id])
+    @actor = Actor.find(params[:actor_id])
+    @movie.actors << @actor
+    redirect_to "/movies/#{@movie.id}"
   end
 
 end
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -1,0 +1,9 @@
+class MoviesController < ApplicationController
+
+  def show
+    @movie = Movie.find(params[:id])
+    @actor_names = @movie.order_by_age
+    @avg_age = @movie.avg_age
+  end
+
+end
--- a/app/models/actor.rb
+++ b/app/models/actor.rb
@@ -1,0 +1,3 @@
+class Actor < ApplicationRecord
+  has_many :movies
+end
--- a/app/models/actor.rb
+++ b/app/models/actor.rb
@@ -1,3 +1,6 @@
 class Actor < ApplicationRecord
-  has_many :movies
+  has_many :movie_actors
+  has_many :movies, through: :movie_actors
+
+  
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -1,3 +1,5 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_many :movie_actors
+  has_many :actors, through: :movie_actors
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -2,4 +2,12 @@ class Movie < ApplicationRecord
   belongs_to :studio
   has_many :movie_actors
   has_many :actors, through: :movie_actors
+
+  def order_by_age
+    actors.order(age: :asc).pluck(:name)
+  end
+
+  def avg_age
+    actors.average(:age).to_f.round(2)
+  end
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -10,4 +10,8 @@ class Movie < ApplicationRecord
   def avg_age
     actors.average(:age).to_f.round(2)
   end
+
+  def self.search(search_params)
+    where("name ILIKE ?", "%#{search_params}%")
+  end
 end
--- a/app/views/movies/show.html.erb
+++ b/app/views/movies/show.html.erb
@@ -4,3 +4,9 @@
 <h3> Genre: <%= @movie.genre %></h3>
 <h5> Actors: <%= @movie.order_by_age %></h5>
 <h5> Actors Average Age: <%= @movie.avg_age %></h5>
+
+  <%= form_with url: "/movies/#{@movie.id}", method: :patch, data: {turbo: false} do |form| %>
+    <p><%= form.label :search_by_name, "Add an actor to this movie" %></p>
+    <%= form.text_field :search_by_name, value: params[:search_by_name] %>
+    <%= form.submit 'submit' %>
+  <% end %>
--- a/app/views/movies/show.html.erb
+++ b/app/views/movies/show.html.erb
@@ -1,0 +1,6 @@
+<h3> Movie Information</h3>
+<h3> Title: <%= @movie.title %></h3>
+<h3> Year: <%= @movie.creation_year %></h3>
+<h3> Genre: <%= @movie.genre %></h3>
+<h5> Actors: <%= @movie.order_by_age %></h5>
+<h5> Actors Average Age: <%= @movie.avg_age %></h5>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -3,4 +3,6 @@ Rails.application.routes.draw do
 
   # Defines the root path route ("/")
   # root "articles#index"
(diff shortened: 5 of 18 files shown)


db/seeds.rb
```diff
@@ -5,3 +5,21 @@
 #
 #   movies = Movie.create([{ name: "Star Wars" }, { name: "Lord of the Rings" }])
 #   Character.create(name: "Luke", movie: movies.first)
+
+@studio1 = Studio.create(name: "Universal Studios", location: "Hollywood")
+@studio2 = Studio.create(name: "Nathan's Studios", location: "Denver")
+
+@movie1 = @studio1.movies.create(title: 'Toy Story', creation_year: '1995', genre: 'animation')
+@movie2 = @studio1.movies.create(title: 'Elephant Christmas', creation_year: '1997', genre: 'romance')
+@movie3 = @studio2.movies.create(title: 'Halloween Boy', creation_year: '2023', genre: 'comedy')
```
juliet-e, line 14: I'm seeing these movies on the studio index page after running your code! Might be worth checking again after you make sure you seed the database

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,20 @@
+class MoviesController < ApplicationController
+
+  def show
+    @movie = Movie.find(params[:id])
+    @actor_names = @movie.order_by_age
```
juliet-e, line 5: It's okay to call @movie.order_by_age in the view so the instance vars on line 5 and 6 are not strictly necessary!

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+    @avg_age = @movie.avg_age
+    if params[:search_by_name].present?
```
juliet-e, line 7: You made this US a bit more complicated (and more similar to the US from the pair project) -- you actually don't need to search for an actor and show the possible actors someone could add. Instead, you just needed to create a form that took an actor ID and routed to a controller action that would associate that actor with the given movie. No need for search -- the user just had to know the actor ID it was interested in adding

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+    @movie = Movie.find(params[:id])
+    @actor_names = @movie.order_by_age
+    @avg_age = @movie.avg_age
+    if params[:search_by_name].present? 
+      threshold = params[:search_by_name].to_s
+      @actors = Actor.filter_by_name(threshold)
+    end
+  end
+
+  def update
```
juliet-e, line 13: Your implementation here is good but a more RESTful option for this action would be a POST because you are creating an association between 2 tables (i.e. you're creating a joins record by creating a MovieActor). 

One RESTful option would be to create a joins controller called the MovieActors controller, and have the form in the view direct to a #create action there. The create action could have this exact logic

app/models/actor.rb
```diff
@@ -0,0 +1,6 @@
+class Actor < ApplicationRecord
+  has_many :movie_actors
```
juliet-e, line 2: Nice M2M Set up!

app/models/movie.rb
```diff
@@ -1,3 +1,17 @@
...
+
+  def order_by_age
+    actors.order(age: :asc).pluck(:name)
+  end
+
+  def avg_age
+    actors.average(:age).to_f.round(2)
+  end
+
+  def self.search(search_params)
```
juliet-e, line 14: No need for search!

app/models/movie.rb
```diff
@@ -1,3 +1,17 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_many :movie_actors
+  has_many :actors, through: :movie_actors
+
+  def order_by_age
```
juliet-e, line 6: Nice work encapsulating this logic into model methods! Very MVC compliant

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,12 @@
+<h3> Movie Information</h3>
+<h3> Title: <%= @movie.title %></h3>
+<h3> Year: <%= @movie.creation_year %></h3>
+<h3> Genre: <%= @movie.genre %></h3>
+<h5> Actors: <%= @movie.order_by_age %></h5>
+<h5> Actors Average Age: <%= @movie.avg_age %></h5>
+
+  <%= form_with url: "/movies/#{@movie.id}", method: :patch, data: {turbo: false} do |form| %>
```
juliet-e, line 8: see note in update action around a more RESTful route. The path for this RESTful route could be `"/movies/:movie_id/actors"`

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,12 @@
...
+    <p><%= form.label :search_by_name, "Add an actor to this movie" %></p>
+    <%= form.text_field :search_by_name, value: params[:search_by_name] %>
```
juliet-e, line 10: Instead of having a name field, the US was looking for a field where a user could enter the actor ID. Then, the ID would be passed to the controller action and could be used to create a MovieActor record

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+    @movie = Movie.find(params[:id])
+    @actor = Actor.find(params[:actor_id])
```
juliet-e, line 15: Your form implementation didn't pass an `actor_id` because your form was looking for a name. So this line probably couldn't execute properly. Your form should have passed the actor ID via the form field

config/routes.rb
```diff
@@ -3,4 +3,9 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/studios", to: "studios#index"
+
+  get "/movies/:id", to: "movies#show"
+  patch "/movies/:id", to: "movies#update"
```
juliet-e, line 10: Consider the more RESTful option: `post "/movies/:id/actors", to: "movie_actors#create"`

spec/features/movie_show_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+
+    end
+    it "has list of actors (name and age) ordered from youngest to oldest" do
+      visit "/movies/#{@movie3.id}"
+
+      expect(page).to have_content(@actor1.name)
+      expect(page).to have_content(@actor2.name)
+      expect(page).to have_content(@actor3.name)
+
+      expect(@actor1.name).to appear_before(@actor2.name)
```
juliet-e, line 45: Nice work using orderly here to test sorting

spec/features/movie_show_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+      expect(@actor2.name).to appear_before(@actor3.name)
+      expect(@actor1.name).to appear_before(@actor3.name)
+    end
+    it "has the average age of all actors in the movie" do
+      visit "/movies/#{@movie3.id}"
+
+      expect(page).to have_content(72.67)
```
juliet-e, line 52: It would be great to anchor this to a text label so that you're not just asserting for a integer, i.e. assert like this:
`expect(page).to have_content("Average Age:  72.67")`

spec/features/studio_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  describe "studio index page" do
+    it "shows name and location as well as movie title, year, and genre" do
+    # User Story 1
+    # As a user,
+    # When I visit the studio index page,
+    # Then I see all of the studios including name and location,
+    # And under each studio I see all of the studio's movies
+    # including the movie's title, creation year, and genre
+      visit "/studios"
+      expect(page).to have_content(@studio1.name)
```
juliet-e, line 21: We should expect to see studio 2 info and movies as well! Make sure to verify the presence of all the studio data (using within blocks would be great here as well)

spec/features/studio_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+      expect(page).to have_content(@studio1.location)
+      expect(page).to have_content(@movie1.title)
+      expect(page).to have_content(@movie1.creation_year)
+      expect(page).to have_content(@movie1.genre)
+    end
+  end
+end
```
juliet-e, line 28: Don't forget to test your model methods in the model spec! Please test associations for your models as well using shoulda matchers

PR #16, opened 2023-10-30
Victor Sanchez- I did better than before :) - finished User Story#1

vsanantone:


app/models/movie_actor.rb
```diff
@@ -0,0 +1,4 @@
+class MovieActor < ApplicationRecord
+  belongs_to :movie
+  belongs_to :actor
```
juliet-e, line 3: Nice many to many set up!

app/views/studios/index.html.erb
```diff
@@ -0,0 +1,14 @@
+<h3>Studios</h3>
+
+<% @studios.each do |studio| %>
+  <section id=<%= "studio-#{studio.id}" %>>
```
juliet-e, line 4: Nice set up with the section ID to allow for robust testing! If, on an IC, you ever get hung up on the within block syntax, feel free to implement the story/test it without within blocks and come back to it at the end if you have remaining time. Looks like you got this working, but I know some students had trouble with the syntax for within blocks with this IC!

spec/features/studios/index_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    star_wars = disney.movies.create!(title: "A New Hope", creation_year: "1979", genre: "sci-fi")
+    avengers = disney.movies.create!(title: "End Game", creation_year: "2019", genre: "action")
+        
+    ice_age = fox_studios.movies.create!(title: "Ice Age", creation_year: "2006", genre: "animated")
+    avatar = fox_studios.movies.create!(title: "Avatar", creation_year: "2009", genre: "sci-fi")
+    # As a user,
+    # When I visit the studio index page,
+    visit "/studios"
+    # Then I see all of the studios including name and location,
+    expect(page).to have_content("Universal Studios")
```
juliet-e, line 21: These assertions could also live inside the within blocks below in order to verify that the location corresponds to the correct studio!

spec/models/actor_spec.rb
```diff
@@ -0,0 +1,7 @@
+require "rails_helper"
+
+RSpec.describe Actor, type: :model do
+  describe "relationships" do
+    it {should have_many :movies}
```
juliet-e, line 5: Please also include a test for the other association through the joins record like so:
`it { should have_many :movie_actors }`
`it { should have_many(:movies).through(:movie_actors) }`

spec/models/movie_spec.rb
```diff
@@ -3,5 +3,7 @@
 RSpec.describe Movie, type: :model do
   describe "relationships" do
     it {should belong_to :studio}
+    it { should have_many :actors }
```
juliet-e, line 6: Please also include a test for the other association through the joins record like so:
`it { should have_many :movie_actors }`
`it { should have_many(:actors).through(:movie_actors) }`

PR #17, opened 2023-10-30
Jimmy Trulock

JimmyTrulock:
Was able to finish up to story 2.

I ran into some syntax issues with the many to many table that ate up some time before I fixed it. 
I also could not get a method to work outside the controller, so I ended up leaving it in the controller. Can you give me some tips on how I could have gotten it to function in the model?
I also ran into an issue with the appears_before not working in my test. For some reason I thought that was capybara not orderly so it took some time and googling to figure that out.
outside that, If I had another hour, I think I would have finished this.


app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,18 @@
+class MoviesController < ApplicationController
+  def show
+    @movie = Movie.find(params[:id])
+    @actors = @movie.actors.order(age: :asc)
+    
+    def average_age(actors)
```
juliet-e, line 6: This method might have given you trouble because you didn't add an `end` for the `show` method before

What we would like to see is this logic encapsulated in a model method (see commend in the Movie model)

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,18 @@
+class MoviesController < ApplicationController
+  def show
+    @movie = Movie.find(params[:id])
+    @actors = @movie.actors.order(age: :asc)
```
juliet-e, line 4: We don't want to do .order in the controller as it is a more complex data operation besides simple CRUD. we would have liked to see it in a model method (see comment in Movie model)

app/models/actor.rb
```diff
@@ -0,0 +1,13 @@
+class Actor < ApplicationRecord
+  has_many :actor_movies
```
juliet-e, line 2: Nice M2M set up! Syntax errors with db associations/setup can be hard so I suggest keeping a cheatsheet with what you did to get this working because it looks good!

app/models/movie.rb
```diff
@@ -1,3 +1,14 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_many :actor_movies
+  has_many :actors, through: :actor_movies
+
+  def average_age(actors)
```
juliet-e, line 6: No need to use Ruby here -- it's much easier in AR!
`actors.average(:age)`

app/models/movie.rb
```diff
@@ -1,3 +1,14 @@
...
+    age = 0
+    actors.each do |actor|
+      age += actor.age
+    end
+    age = age / actors.count
+    age
+  end
```
juliet-e, line 13: We'd also like to see a method here (instead of controller) to order the actors:
```
def order_actors_youngest_first
    actors.order(age: :asc)
end
```

juliet-e: then, in the view, you can call this on a movie instance and iterate over the actors: 
`@movie.order_actors_ youngest_first.each do |actor|`

config/routes.rb
```diff
@@ -3,4 +3,7 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+  get "/studios", to: "studios#index"
+  get "movies/:id", to: "movies#show"
```
juliet-e, line 7: make sure to start every path with a `/`

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+      @movie_1 = @studio_1.movies.create(title: "Star Wars: The Last Jedi", creation_year: "2017", genre: "Sci-fi")
+      @actor_1 = Actor.create(name: "Mark Hamill", age: 72)
+      @actor_2 = Actor.create(name: "Adam Driver", age: 39)
+      ActorMovie.create(movie: @movie_1, actor: @actor_1)
+      ActorMovie.create(movie: @movie_1, actor: @actor_2)
+
+      visit "/movies/#{@movie_1.id}"
+
+      expect(page).to have_content("Star Wars: The Last Jedi")
+      expect(page).to have_content("2017")
```
juliet-e, line 16: It's more meaningful to have this data anchored to what it represents, consider an assertion like this:
`expect(page).to have_content("Date Created: 2017")`

spec/features/studios/index_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+      @studio_2 = Studio.create(name: "Universal", location: "Hollywood")
+      @movie_1 = @studio_1.movies.create(title: "Star Wars: The Last Jedi", creation_year: "2017", genre: "Sci-fi")
+      @movie_2 = @studio_2.movies.create(title: "Batman: The Killing Joke", creation_year: "2016", genre: "Animated")
+      
+      visit "/studios"
+
+      expect(page).to have_content("Disney")
+      expect(page).to have_content("Universal")
+      expect(page).to have_content("Star Wars: The Last Jedi")
+      expect(page).to have_content("Batman: The Killing Joke")
```
juliet-e, line 16: This test could be more robust if you used within blocks to section off different studios. Right now, you cannot verify if Star Wars is under Disney or Universal

spec/models/movie_spec.rb
```diff
@@ -3,5 +3,7 @@
 RSpec.describe Movie, type: :model do
   describe "relationships" do
     it {should belong_to :studio}
+    it {should have_many :actor_movies}
+    it {should have_many :actors}
```
juliet-e, line 7: Make sure to test your model methods!

PR #18, opened 2023-10-30
Miranda Delaney

delaneymiranda1:
I felt this one was a little harder to get through. I spent a decent amount of time creating migrations and test data and I only partially finished user story 3.

### DIFF
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -1,0 +1,7 @@
+class MoviesController < ApplicationController
+  def show
+    @movie = Movie.includes(:actors).find(params[:id])
+    @actors = @movie.actors.order(age: :asc)
+    @average_age = @actors.average(:age)
+  end
+end
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -4,4 +4,16 @@ class MoviesController < ApplicationController
     @actors = @movie.actors.order(age: :asc)
     @average_age = @actors.average(:age)
   end
+
+  def update
+    @movie = Movie.find(params[:id])
+    
+    redirect_to @movie
+  end
+  
+  private
+  
+  def movie_params
+    params.require(:movie).permit(:actor_id)
+  end
 end
--- a/app/models/actor.rb
+++ b/app/models/actor.rb
@@ -1,0 +1,3 @@
+class Actor < ApplicationRecord
+  has_many :movies, through: :studio_movies
+end
--- a/app/models/actor.rb
+++ b/app/models/actor.rb
@@ -1,3 +1,3 @@
 class Actor < ApplicationRecord
-  has_many :movies, through: :studio_movies
+  has_and_belongs_to_many :movies
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -1,3 +1,4 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_many :actors
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -1,4 +1,5 @@
 class Movie < ApplicationRecord
   belongs_to :studio
-  has_many :actors
+  has_and_belongs_to_many :actors
+  # has_many :actors, through: :actor_movie
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -1,5 +1,9 @@
 class Movie < ApplicationRecord
   belongs_to :studio
   has_and_belongs_to_many :actors
-  # has_many :actors, through: :actor_movie
+  
+  def add_actor(actor_id)
+    actor = Actor.find(actor_id)
+    self.actors << actor unless self.actors.include?(actor)
+  end
 end
--- a/app/views/movies/show.html.erb
+++ b/app/views/movies/show.html.erb
@@ -1,0 +1,12 @@
+<h1><%= @movie.title %></h1>
+<p>Year: <%= @movie.creation_year %></p>
+<p>Genre: <%= @movie.genre %></p>
+
+<h2>Actors</h2>
+<ul>
+  <% @actors.each do |actor| %>
+    <li><%= actor.name %> - Age: <%= actor.age %></li>
+  <% end %>
+</ul>
+
+<p>Average Age of Actors: <%= @average_age %></p>
--- a/app/views/movies/show.html.erb
+++ b/app/views/movies/show.html.erb
@@ -10,3 +10,9 @@
 </ul>
 
 <p>Average Age of Actors: <%= @average_age %></p>
+
+<p><%= form_for @movies do |f| %>
+  <%= f.label :actor_id, "Add Actor by ID:" %>
+  <%= f.text_field :actor_id %>
+  <%= f.submit "Add Actor" %>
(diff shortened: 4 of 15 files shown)


app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,19 @@
+class MoviesController < ApplicationController
+  def show
+    @movie = Movie.includes(:actors).find(params[:id])
+    @actors = @movie.actors.order(age: :asc)
```
juliet-e, line 4: Remember that we don't want to use `.order` or `.average` in the controller since they are AR methods beyond simple CRUD functions. These are MVC infractions as we want to be mindful of not putting too much complexity in controllers, especially when it comes to database operations.

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+    @average_age = @actors.average(:age)
+  end
+
+  def update
```
juliet-e, line 8: To add an actor to a movie, we are actually creating an joins record that associates an existing movie to an existing actor. In this sense, we're actually *creating* something (an ActorMovie) rather than *updating* the movie resources. So, a more RESTful option here would be to route the form in your view to a `POST` to `/movies/:movie_id/actors` and that can direct to a joins controller: `ActorMoviesController#create`

In this create action, you could find the existing movie and actor from the params, and then create a ActorMovie object, and redirect back to the movie show page.

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+    @movie = Movie.find(params[:id])
+    
+    redirect_to @movie
+  end
+  
+  private
+  
+  def movie_params
+    params.require(:movie).permit(:actor_id)
```
juliet-e, line 17: The form syntax you're using in the view is a bit old -- consider form_with url and then you won't have to use the `require(:movie)` and can just permit :actor_id

app/models/actor.rb
```diff
@@ -0,0 +1,3 @@
+class Actor < ApplicationRecord
+  has_and_belongs_to_many :movies
```
juliet-e, line 2: This association doesn't always work with the many to many relationships we use in Mod 2 so we discourage folks from using it. Instead, use 
```
has_many :actor_movies
has_many :movies, through: :actor_movies
```
We don't like using has_and_belongs_to_many because it suggests that your join table can *only* have foreign keys and nothing else. There are times in our projects that your joins tables also have other attributes so we encourage learning this syntax instead! In this case, it's still going to work because we don't have attributes besides foreign keys in this joins table, but on projects, it might cause problems. This doesn't penalize you, just want you to be aware.

app/models/movie.rb
```diff
@@ -1,3 +1,9 @@
 class Movie < ApplicationRecord
```
juliet-e, line 1: We'd want the ordering and averaging methods to live here!

app/models/movie.rb
```diff
@@ -1,3 +1,9 @@
...
   belongs_to :studio
+  has_and_belongs_to_many :actors
+  
+  def add_actor(actor_id)
+    actor = Actor.find(actor_id)
+    self.actors << actor unless self.actors.include?(actor)
```
juliet-e, line 7: This is okay to put in a model method but just finding an actor and adding it to a movie is simple CRUD logic that could live in a controller (as opposed to the more complex querying calculations with average, order, etc which we want to live in a model method)

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,18 @@
...
+<h2>Actors</h2>
+<ul>
+  <% @actors.each do |actor| %>
+    <li><%= actor.name %> - Age: <%= actor.age %></li>
+  <% end %>
+</ul>
+
+<p>Average Age of Actors: <%= @average_age %></p>
+
+<p><%= form_for @movies do |f| %>
```
juliet-e, line 14: `form_for` is deprecated, so you should use `form_with url` in this case!

config/routes.rb
```diff
@@ -3,4 +3,7 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/studios", to: "studios#index"
+  get "/movies/:id", to: "movies#show"
```
juliet-e, line 8: You'll need a route here for submitting the form. A RESTful option could be:
`post "/movies/:id/actors", to: "actor_movies#create"`

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    ActorMovie.create(actor_id: @billy_eichner.id, movie_id: @bros.id)
+    ActorMovie.create(actor_id: @luke_macfarlane.id, movie_id: @bros.id)
+    @bros.actors << [@billy_eichner, @luke_macfarlane]
+  end
+
+  describe "When I visit a movie's show page" do
+    it 'displays movies attributes and all actors youngest to oldest in movie, with average age' do
+      visit "/movies/#{@bros.id}"
+      expect(page).to have_content("Bros")
+      expect(page).to have_content("2022")
```
juliet-e, line 30: It's best to anchor this data to labels that describe what it refers to, and then test for the whole text. For example, expecting the page to have `"Bros, Released in 2022"` is more robust than expecting just `Bros` and `2022` as separate assertions. That text could be anywhere so it's more powerful to anchor these fields together. Even more robust would be to add within blocks to assert that this data appears in the section for Bros and not elsewhere

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      expect(page).to have_content("romcom")
+      expect(page).to have_content("Billy Eichner")
+      expect(page).to have_content("45")
+      expect(page).to have_content("Luke Macfarlane")
+      expect(page).to have_content("43")
+      
+      expect(page).to have_text(/.*Luke Macfarlane.*Billy Eichner.*/)
+      expect(page).to have_content("Average Age of Actors: 44")
```
juliet-e, line 38: This is a more robust assertion!

spec/features/studios/index_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      expect(page).to have_content("2023")
+      expect(page).to have_content("horror")
+      expect(page).to have_content("Bros")
+      expect(page).to have_content("2022")
+      expect(page).to have_content("romcom")
+      expect(page).to have_content("Paramount")
+      expect(page).to have_content("Los Angeles")
+      expect(page).to have_content("Hot Rod")
+      expect(page).to have_content("2007")
+      expect(page).to have_content("comedy")
```
juliet-e, line 36: Same deal here - These assertions are not as significant when they are not grouped together or anchored to eachother. How do we know that 2007 refers to Hot Rod and not M3GAN? How do we know that it's not part of the footer of a page where there's a phone number that just happens to include 2007? This test can more robustly verify behavior when you anchor text together and use within blocks

spec/features/studios/index_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+  end
+end
```
juliet-e, line 39: Don't forget to test your model associations in model specs! Once encpauslating data operations in model methods, we'd love to see those tested in model tests as well

PR #19, opened 2023-10-30
Matt Darlington

mdarl17:
Made a few mistakes that cost me a lot of time.  Definitely harder than first one. Disappointed I didn't do better (at minimum finishing all testing/functionality through User Story #3)

### DIFF
--- a/app/controllers/movie_actors_controller.rb
+++ b/app/controllers/movie_actors_controller.rb
@@ -1,3 +1,13 @@
 class MovieActorsController < ApplicationController
-  
+  def new
+
+  end
+
+  def create
+    movie = Movie.find(params[:id])
+    actor = movie.actors.find(params[:actor_id])
+    @movie.actors.create!({movie_id: movie.id, actor_id: actor.id})
+
+    redirect_to "/movies/#{movie.id}"
+  end
 end
--- a/app/controllers/movie_actors_controller.rb
+++ b/app/controllers/movie_actors_controller.rb
@@ -5,8 +5,8 @@ class MovieActorsController < ApplicationController
 
   def create
     movie = Movie.find(params[:id])
-    actor = movie.actors.find(params[:actor_id])
-    @movie.actors.create!({movie_id: movie.id, actor_id: actor.id})
+    actor = Actor.find(params[:actor_id])
+    MovieActor.create!({actor_id: actor.id, movie_id: movie.id})
 
     redirect_to "/movies/#{movie.id}"
   end
--- a/app/controllers/movie_actors_controller.rb
+++ b/app/controllers/movie_actors_controller.rb
@@ -1,0 +1,3 @@
+class MovieActorsController < ApplicationController
+  
+end
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -1,0 +1,9 @@
+class MoviesController < ApplicationController
+
+  def show
+    @movie = Movie.find(params[:id])
+    @actors = @movie.actor_sort
+    @avg_age = @movie.avg_age
+  end
+  
+end
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -5,10 +5,5 @@ class MoviesController < ApplicationController
     @actors = @movie.actor_sort
     @avg_age = @movie.avg_age
   end
-
-  def new
-
-  end
-  
   
 end
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -5,5 +5,10 @@ class MoviesController < ApplicationController
     @actors = @movie.actor_sort
     @avg_age = @movie.avg_age
   end
+
+  def new
+
+  end
+  
   
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -1,3 +1,13 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_many :movie_actors
+  has_many :actors, through: :movie_actors
+
+  def actor_sort
+    actors.order(age: :asc)
+  end
+
+  def avg_age
+    actors.average(:age).to_f.round(1)
+  end
 end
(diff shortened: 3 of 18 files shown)


mdarl17:
Made a few mistakes that cost me a lot of time.  Definitely harder than first one. Disappointed I didn't do better (at minimum finishing all testing/functionality through User Story #3)

app/controllers/movie_actors_controller.rb
```diff
@@ -0,0 +1,13 @@
+class MovieActorsController < ApplicationController
+  def new
+
+  end
+
+  def create
+    movie = Movie.find(params[:id])
```
juliet-e, line 7: Excellent controller organization with this joins controller!

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,9 @@
+class MoviesController < ApplicationController
+
+  def show
+    @movie = Movie.find(params[:id])
+    @actors = @movie.actor_sort
```
juliet-e, line 5: Having the @actors and @avg_age instance variables is not strictly necessary because you can do @movie.actor_sort and .avg_age from the view -- just FYI! This is not wrong but we generally try to minimize the number of instance variables we set up in controllers

app/models/movie.rb
```diff
@@ -1,3 +1,13 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_many :movie_actors
+  has_many :actors, through: :movie_actors
+
+  def actor_sort
```
juliet-e, line 6: Excellent work encapsulating this into model methods!

config/routes.rb
```diff
@@ -3,4 +3,11 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/studios", to: "studios#index"
+
+  get "/movies/:id", to: "movies#show"
+  get "/movies/:id/movie_actors/new", to: "movie_actors#new"
```
juliet-e, line 10: This /new path is actually not necessary because we're rendering the form on the show page. We only need a new page if we're rendering the form to create a new resource on a separate page

config/routes.rb
```diff
@@ -3,4 +3,11 @@
...
+  post "movies/:id/movie_actors", to: "movie_actors#create"
+
```
juliet-e, line 12: Great RESTful routes!

spec/models/movie_spec.rb
```diff
@@ -4,4 +4,35 @@
   describe "relationships" do
     it {should belong_to :studio}
   end
+
```
juliet-e, line 7: Your model tests for these instance methods look good. Don't forget to test the relationships though - we want to see here, for example:
```
it {should have_many :movie_actors}
it {should have_many(:actors).through(:movie_actors)}
```

We should have association tests in the actor spec and movie_actor spec as well

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+    robert = Actor.create!({name: "Robert Shaw", age: 61})
+
+    jaws_actor_1 = MovieActor.create!({actor_id: roy.id, movie_id: jaws.id})
+    jaws_actor_2 = MovieActor.create!({actor_id: richard.id, movie_id: jaws.id})
+    jaws_actor_3 = MovieActor.create!({actor_id: robert.id, movie_id: jaws.id})
+
+    visit "/movies/#{jaws.id}"
+
+    expect(page).to have_content("Cast:")
+    expect(page).to have_content("Richard Dreyfuss, age: 28")
```
juliet-e, line 32: you can add the orderly gem here to test that Richard Dreyfuss appears before Roy Schneider, etc

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+    jaws = paramount.movies.create({title: "Jaws", creation_year: "1976", genre: "Action"})
+
+    visit "/movies/#{jaws.id}"
+
+    expect(page).to have_content("Title: Jaws")
+    expect(page).to have_content("Year: 1976")
+    expect(page).to have_content("Genre: Action")
+  end
+
+  it "shows each movie's actors sorted from youngest to oldest" do
```
juliet-e, line 16: This test case can just be edited to "shows each movie's actors" since it's not actually testing sorting here (that's below)

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+      visit "/movies/#{jaws.id}"
+
+      expect("Richard Dreyfuss").to appear_before("Roy Scheider")
+      expect("Roy Scheider").to appear_before("Robert Shaw")
+
+    end
+
+    it "has a form to add an actor to a movie" do
+
+      paramount = Studio.create!({name: "Paramount Pictures", location: "Hollywood"})
```
juliet-e, line 83: Consider using a `before :each` since you are repeating test data!

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+      expect(page).to have_content("Add Actor")
+      expect(page).to have_field(:actor_id)
+      expect(page).to have_button("Add Actor")
+
+      fill_in(:actor_id, with: lorraine.id)
+
+      click_button("Add Actor")
+
+      expect(current_path).to eq("/movies/#{jaws.id}")
+      expect(page).to have_content("Lorraine Gary, age: 38")
```
juliet-e, line 106: Robust testing! Great work

spec/features/studios/index_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    mgm = Studio.create!({name: "MGM", location: "Burbank"})
+    universal = Studio.create!({name: "Universal Studios", location: "Hollywood"})
+
+    jaws = paramount.movies.create({title: "Jaws", creation_year: "1976", genre: "Action"})
+    kane = mgm.movies.create({title: "Citizen Kane", creation_year: "1942", genre: "Drama"})
+    kong = universal.movies.create({title: "King Kong", creation_year: "1936", genre: "Action"})
+
+    visit "/studios"
+    
+    expect(page).to have_content("Title: Jaws")
```
juliet-e, line 37: Because the order or organization of our assertions don't matter, the only way to determine if "Year: 1976" is under Jaws is to use a within block.

PR #20, opened 2023-10-30
TP Pubins

trpubz:
I felt like I killed it. Thought maybe could use `render` on the form action but couldn’t get it

### DIFF
--- a/app/controllers/movie_actors_controller.rb
+++ b/app/controllers/movie_actors_controller.rb
@@ -1,0 +1,7 @@
+class MovieActorsController < ApplicationController
+  def create
+    MovieActor.create!(movie_id: params[:id], actor_id: params[:actor_id])
+
+    redirect_to "/movies/#{params[:id]}"
+  end
+end
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -1,0 +1,5 @@
+class MoviesController < ApplicationController
+  def show
+    @movie = Movie.find(params[:id])
+  end
+end
--- a/app/controllers/studios_controller.rb
+++ b/app/controllers/studios_controller.rb
@@ -1,0 +1,5 @@
+class StudiosController < ApplicationController
+  def index
+    @studios = Studio.all
+  end
+end
--- a/app/controllers/studios_controller.rb
+++ b/app/controllers/studios_controller.rb
@@ -2,4 +2,8 @@ class StudiosController < ApplicationController
   def index
     @studios = Studio.all
   end
+
+  def show
+    @studio = Studio.find(params[:id])
+  end
 end
--- a/app/models/actor.rb
+++ b/app/models/actor.rb
@@ -1,0 +1,12 @@
+class Actor < ApplicationRecord
+  has_many :movie_actors
+  has_many :movies, through: :movie_actors
+
+  def self.youngest_to_oldest
+    order(age: :asc)
+  end
+
+  def self.average_age
+    average(:age)
+  end
+end
--- a/app/models/studio.rb
+++ b/app/models/studio.rb
@@ -1,3 +1,9 @@
 class Studio < ApplicationRecord
   has_many :movies
+  has_many :movie_actors, through: :movies
+  has_many :actors, through: :movie_actors
+
+  def unique_actors
+    actors.distinct
+  end
 end
--- a/app/views/movies/show.html.erb
+++ b/app/views/movies/show.html.erb
@@ -1,0 +1,16 @@
+<h1><%= @movie.title %></h1>
+<h2>Year: <%= @movie.creation_year %></h2>
+<h2>Genre: <%= @movie.genre %></h2>
+
+<h3>Actors</h3>
+<% @movie.actors.youngest_to_oldest.each do | actor | %>
+    <p><%= actor.name %></p>
+<% end %>
+<p>Average Age: <%= @movie.actors.average_age.round(1) %></p>
+
+<%= form_with url: "/movies/#{@movie.id}/actors", method: :post, data: {turbo: false} do |f| %>
+  <%= f.label :actor_id, "Add Actor by ID" %>
+  <%= f.text_field :actor_id %>
+
+  <%= f.submit "Add Actor" %>
+<% end %>
--- a/app/views/studios/index.html.erb
+++ b/app/views/studios/index.html.erb
@@ -1,0 +1,8 @@
+<h1>Studios</h1>
+
+<% @studios.each do | studio | %>
+    <h2><%= "#{studio.name}, #{studio.location}" %></h2>
+    <% studio.movies.each do | movie | %>
+        <h3><%= "#{movie.title} | Year: #{movie.creation_year} | #{movie.genre}" %></h3>
+    <% end %>
+<% end %>
--- a/app/views/studios/show.html.erb
+++ b/app/views/studios/show.html.erb
@@ -1,0 +1,7 @@
+<h1><%= @studio.name %></h1>
+<h2><%= @studio.location %></h2>
+
+<h3>Studio Actors</h3>
+<% @studio.unique_actors.each do | actor | %>
+    <p><%= actor.name %></p>
+<% end %>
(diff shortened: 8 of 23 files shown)


app/controllers/movie_actors_controller.rb
```diff
@@ -0,0 +1,7 @@
+class MovieActorsController < ApplicationController
+  def create
+    MovieActor.create!(movie_id: params[:id], actor_id: params[:actor_id])
+
+    redirect_to "/movies/#{params[:id]}"
+  end
```
abdulredd, line 6: Perfect. Great use of the join controller to make the connection between an actor and a movie!

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,5 @@
+class MoviesController < ApplicationController
+  def show
+    @movie = Movie.find(params[:id])
+  end
+end
```
abdulredd, line 5: 👍

app/controllers/studios_controller.rb
```diff
@@ -0,0 +1,9 @@
+class StudiosController < ApplicationController
+  def index
+    @studios = Studio.all
+  end
+
+  def show
+    @studio = Studio.find(params[:id])
+  end
+end
```
abdulredd, line 9: 👍

app/models/actor.rb
```diff
@@ -0,0 +1,12 @@
...
+  has_many :movies, through: :movie_actors
+
+  def self.youngest_to_oldest
+    order(age: :asc)
+  end
+
+  def self.average_age
+    average(:age)
+  end
+end
```
abdulredd, line 12: This is quite clever. Using "class" methods on your `Actor` model since you can call class methods on an AR collection. Interesting approach and one that makes a lot of sense. Great implementation. One think I will note is that ascending is the default so `order(:age)` also works. :)

app/models/studio.rb
```diff
@@ -1,3 +1,9 @@
 class Studio < ApplicationRecord
   has_many :movies
+  has_many :movie_actors, through: :movies
+  has_many :actors, through: :movie_actors
+
```
abdulredd, line 5: Not sure what `:movie_actors` is giving you here but my assumption is that you are specifying that association so you can get to `:actors`. If that is the case, you can skip the middle man and just go to actors from movies like so: `has_many :actors, through: :movies` because in your `Movie` model, there's access to `:actors` through `:movie_actors` already.

If this is at all confusing, please don't hesitate to let me know.

app/models/studio.rb
```diff
@@ -1,3 +1,9 @@
...
+  def unique_actors
+    actors.distinct
+  end
```
abdulredd, line 8: 👍

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,16 @@
+<h1><%= @movie.title %></h1>
+<h2>Year: <%= @movie.creation_year %></h2>
+<h2>Genre: <%= @movie.genre %></h2>
+
+<h3>Actors</h3>
+<% @movie.actors.youngest_to_oldest.each do | actor | %>
+    <p><%= actor.name %></p>
+<% end %>
+<p>Average Age: <%= @movie.actors.average_age.round(1) %></p>
```
abdulredd, line 9: I'm pointing this one line out for a reason, this is great! I am in the camp that also considers rounding or any formatting of data to be presentation logic, so kudos here.

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,16 @@
...
+
+<%= form_with url: "/movies/#{@movie.id}/actors", method: :post, data: {turbo: false} do |f| %>
+  <%= f.label :actor_id, "Add Actor by ID" %>
+  <%= f.text_field :actor_id %>
+
+  <%= f.submit "Add Actor" %>
+<% end %>
```
abdulredd, line 16: Yes! This form is perfect and follows restful conventions perfectly! You're using the correct verb (POST) and sending the data to the correct path, great job TP!

app/views/studios/index.html.erb
```diff
@@ -0,0 +1,8 @@
+<h1>Studios</h1>
+
+<% @studios.each do | studio | %>
+    <h2><%= "#{studio.name}, #{studio.location}" %></h2>
+    <% studio.movies.each do | movie | %>
+        <h3><%= "#{movie.title} | Year: #{movie.creation_year} | #{movie.genre}" %></h3>
+    <% end %>
+<% end %>
```
abdulredd, line 8: 👍

app/views/studios/show.html.erb
```diff
@@ -0,0 +1,7 @@
+<h1><%= @studio.name %></h1>
+<h2><%= @studio.location %></h2>
+
+<h3>Studio Actors</h3>
+<% @studio.unique_actors.each do | actor | %>
+    <p><%= actor.name %></p>
+<% end %>
```
abdulredd, line 7: 👍

config/routes.rb
```diff
@@ -3,4 +3,10 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+  get "/studios", to: "studios#index"
+  get "/studios/:id", to: "studios#show"
+
+  get "/movies/:id", to: "movies#show"
+
+  post "/movies/:id/actors", to: "movie_actors#create"
 end
```

db/migrate/20231030153501_create_movie_actors.rb
```diff
@@ -0,0 +1,10 @@
+class CreateMovieActors < ActiveRecord::Migration[7.0]
+  def change
+    create_table :movie_actors do |t|
+      t.references :movie, null: false, foreign_key: true
+      t.references :actor, null: false, foreign_key: true
+
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 10: Great job creating the joins table.

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    # (You do not have to test for a sad path, for example if the id submitted is not an existing actor)
+    visit "movies/#{@spiderman.id}"
+    
+    fill_in "Add Actor by ID", with: @jjohnson.id
+    click_button "Add Actor"
+    
+    expect(page).to have_content @jjohnson.name
+    expect(page).to have_content "Average Age: 33.0"
+  end
+end
```
abdulredd, line 63: Great job with these tests. I only have one suggestion for the future and that is to try and reduce the ambiguity by using `within` blocks or specifying more specific assertions. Here's what I mean by that. This assertion, `expect(page).to have_content @spiderman.creation_year` just says that `2023` just needs to show up on the page somewhere, it could be part of a date, a phone number, etc. It doesn't specify that it's the year of the movie at all.

This could be resolved by placing this inside a `within` block or specifying a more specific assertion like: `expect(page).to have_content("Year: #{@spiderman.creation_year}")`. While this may seem like a small thing, it helps reduce ambiguity and helps avoid false positives by testing exactly what and where you intend to test.

This one is great: `expect(page).to have_content "Average Age: 27.0"`, no ambiguity here.

spec/models/actor_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+    
+    describe "#average_age" do
+      it "averages the ages" do
+        expect(Actor.average_age.to_i).to eq 107/3
+        expect(@spiderman.actors.average_age.to_i).to eq 27
+      end
+    end
+  end
+end
```
abdulredd, line 36: Great job testing all your class methods. This is a small thing, but the syntax for class methods is `::` 😅

`describe "::youngest_to_oldest"`
`describe "::average_age"`

I'm splitting hairs at this point lol.

spec/models/studio_spec.rb
```diff
@@ -3,5 +3,34 @@
...
+      MovieActor.create!(movie_id: @spiderman1.id, actor_id: @jjohnson.id)
+    end
+
+    describe "#unique_actors" do
+      it "shows the distinct actors that have worked for the studio" do
+        expect(@sony.unique_actors).to eq [@smoore, @hsteinfeld, @jjohnson]
+      end
+    end
   end
 end
```
abdulredd, line 36: 👍

PR #21, opened 2023-10-30
Anthea Yur

chisPmama:
Finished Extension + Fun! I had a lot of fun with this! I feel really confident with my work and was able to add some extra flare and functionality with 100% simplecov test coverage. Thanks for posting a practice exam! It helped so much.


app/controllers/movie_actor_controller.rb
```diff
@@ -0,0 +1,9 @@
+class MovieActorController < ApplicationController
```
cjsim89, line 1: Good use of this join controller! It's exactly the approach I like to take too.

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,24 @@
...
+    <%= form.hidden_field :search, value: "actor" %>
+    <%= form.label :actor_id, "Add Existing Actor" %>
+    <%= form.text_field :actor_id %>
+    <%= form.submit "Submit" %>
+  <% end %>
+
+  <% if !@search_actors.nil? && @search_actors.any? %>
+  <% @search_actors.each do |actor| %>
+  <p><b>ID: <%= actor.id %>  |  <%= actor.name%> | Age: <%= actor.age%><br>
+  <%= button_to "Add ##{actor.id}", {controller: "movie_actor", action: "create", movie_id: @movie.id, actor_id: actor.id}, method: :post%>
```
cjsim89, line 22: This works, but the approach is a little bit, uh, "extra" from what the user story is asking for ,which is to enter an existing Actor's ID and add them to the join table. 
That having been said, this is implemented well! Normally we don't need this double-layer approach though.

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,10 @@
+class MoviesController < ApplicationController
+  def show
+    @movie = Movie.find(params[:id])
+    if params[:search] == "actor" && params[:actor_id].present? 
+      actor_id = params[:actor_id]
+      @search_actors = Actor.where("id::text LIKE ?", "%#{actor_id}%")
```
cjsim89, line 6: This would be an MVC infraction because of the use of `.where` - this should be in the model and tested.

app/models/studio.rb
```diff
@@ -1,3 +1,9 @@
 class Studio < ApplicationRecord
   has_many :movies
+
+  def actors_featured
+    movie_ids = movies.pluck(:id)
+    actors_featured = Actor.joins(:movies,:movie_actors).where(movies: {id: movie_ids}).distinct
```
cjsim89, line 6: This works, but we can make it simpler by adding an additional association: `has_many :actors, through: :movies`. 
The reason for this is that a Studio already has many movies, and a movie has many actors.

cjsim89: If we added that, this query becomes

```
actors.distinct
```

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,24 @@
...
+<p><b>Creation Year: </b><%= @movie.creation_year %></p>
+<p><b>Genre: </b><%= @movie.genre %></p>
+<p><b>Actors: </b><br><% @movie.actors.sort_asc_age.each do |a|%>
+                       ID: <%= a.id %>  |  <%= a.name%> | Age: <%= a.age%><br>
+                       <% end%>
+
+<p><b>Average Age of Actors:</b> <%= @movie.average_age %> years old 
+
+<h3>Add an Existing Actor to Movie by ID</h3>
+  <%= form_with url: "/movies/#{@movie.id}", method: :get, data: { turbo: false } do |form| %>
```
cjsim89, line 12: The way you have this set up with the search form + button combo, I agree that this would be a `:get`. If we combined it all into one form, then the method should be `:post`. This is ok for now, but be careful about this kind of approach; I wouldn't want you to lose time in the final or retake is all.

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,24 @@
...
+    <%= form.hidden_field :search, value: "actor" %>
```
cjsim89, line 13: What is the point of this hidden field?

spec/features/studios/show_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    it 'when viewing the show page, there is the studio name and location' do
+      visit "/studios/#{@studioghibli.id}"
+      expect(page).to have_content(@studioghibli.name)
+      expect(page).to have_content(@studioghibli.location)
+    end
+
+    it 'also includes a unique list of all the actors that have been in any of the studio movies' do
+      visit "/studios/#{@studioghibli.id}"
+      studio_actors = []
+      @studioghibli.movies.each{|movie| studio_actors << movie.actors}
```
cjsim89, line 18: Be careful about condensing code like this in a test - it can seem slick at the time, but can become harder to read or to refactor later. Sometimes it is better in the long run to write it all out.

spec/spec_helper.rb
```diff
@@ -13,6 +13,32 @@
...
+  @catbus = Actor.create!(name: "Cat Bus", age: 400)
+  @mike = Actor.create!(name: "Mike Wazowski", age: 40)
+  @baby_totoro = Actor.create!(name: "Baby Totoro", age: 140)
+
+  MovieActor.create!(movie: @spiritedaway, actor: @noface)
+  MovieActor.create!(movie: @spiritedaway, actor: @sootsprite)
+  MovieActor.create!(movie: @totoro, actor: @mini_totoro)
+  MovieActor.create!(movie: @totoro, actor: @sootsprite)
+  MovieActor.create!(movie: @totoro, actor: @catbus)
+  MovieActor.create!(movie: @monstersinc, actor: @mike)
```
cjsim89, line 38: Good test data setup!

cjsim89 (commented):
Overall good work!

PR #22, opened 2023-10-30
Edward Rodriguez

TheAveryRodriguez:
Thought that everything was going well but ended up getting stuck at the end of US 3. Otherwise everything seemed good and I felt confident about the EXT as well.

I think I was second guessing some things and ended up wasting too much time. Got the best of me today but wont hold me back.

### DIFF
--- a/app/controllers/movie_actors_controller.rb
+++ b/app/controllers/movie_actors_controller.rb
@@ -1,0 +1,9 @@
+class MovieActorsController < ApplicationController
+  def create
+    @movie = Movie.find(params[:movie_id])
+    @actor = Actor.find(params[:actor_id])
+
+    @movie.actors << @actor
+    redirect_to "/movies.#{@movie.id}"
+  end
+end
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -1,0 +1,6 @@
+class MoviesController < ApplicationController
+  def show
+    @movie = Movie.find(params[:id])
+    @actors_average_age = @movie.average_age
+  end
+end
--- a/app/models/actor.rb
+++ b/app/models/actor.rb
@@ -1,0 +1,7 @@
+class Actor < ApplicationRecord
+  has_many :movie_actors
+  has_many :movies, through: :movie_actors
+
+  validates :name, presence: true
+  validates :age, presence: true
+end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -1,3 +1,17 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_many :movie_actors
+  has_many :actors, through: :movie_actors
+
+  validates :title, presence: true
+  validates :creation_year, presence: true
+  validates :genre, presence: true
+
+  def youngest_first
+    actors.order(:age)
+  end
+
+  def average_age
+    actors.average(:age)
+  end
 end
--- a/app/views/movies/show.html.erb
+++ b/app/views/movies/show.html.erb
@@ -1,0 +1,23 @@
+<h1>Movie Show Page</h1>
+
+  <h3><br><%= "Title: #{@movie.title}" %></h5>
+  <%= "Creation Year: #{@movie.creation_year}" %>
+  <%= "Genre: #{@movie.genre}" %>
+ 
+  <h4><br><%= "Actors in #{@movie.title}" %></h4>
+  <% @movie.youngest_first.each do |actor| %>
+        <%= "Name: #{actor.name}" %>
+    <br>
+  <% end %>
+
+  <h4><%= "Actors Average Age: #{@actors_average_age}" %></h4>
+<br>
+
+  <section id="add-actor">
+  <h3>Add an actor to this movie</h3>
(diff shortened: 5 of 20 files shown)


app/controllers/movie_actors_controller.rb
```diff
@@ -0,0 +1,9 @@
+class MovieActorsController < ApplicationController
```
juliet-e, line 1: Awesome controller organization! Using the create action in this joins controller is a very RESTful architecture

app/controllers/movie_actors_controller.rb
```diff
@@ -0,0 +1,9 @@
...
+  def create
+    @movie = Movie.find(params[:movie_id])
+    @actor = Actor.find(params[:actor_id])
+
+    @movie.actors << @actor
+    redirect_to "/movies.#{@movie.id}"
```
juliet-e, line 7: I think you  mean "/movies/#{movie.id}" instead of using `.`!

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,6 @@
+class MoviesController < ApplicationController
+  def show
+    @movie = Movie.find(params[:id])
+    @actors_average_age = @movie.average_age
```
juliet-e, line 4: Just for the future -- this second instance var is not strictly necessary because you can call `@movie.average_age` in the view, too! Calling a model method from the view is just fine, we just don't want to call AR methods from the view

app/models/actor.rb
```diff
@@ -0,0 +1,7 @@
+class Actor < ApplicationRecord
+  has_many :movie_actors
```
juliet-e, line 2: M2M set up looks good!

app/models/movie.rb
```diff
@@ -1,3 +1,17 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_many :movie_actors
+  has_many :actors, through: :movie_actors
+
+  validates :title, presence: true
+  validates :creation_year, presence: true
+  validates :genre, presence: true
+
+  def youngest_first
```
juliet-e, line 10: Excellent MVC-compliance with the encapsulation this data logic

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,23 @@
...
+        <%= "Name: #{actor.name}" %>
+    <br>
+  <% end %>
+
+  <h4><%= "Actors Average Age: #{@actors_average_age}" %></h4>
+<br>
+
+  <section id="add-actor">
+  <h3>Add an actor to this movie</h3>
+  <%= form_with url: "/movie_actors", method: :post, data: { turbo: false } do |form| %>
```
juliet-e, line 18: Awesome RESTful route here! You might have run into trouble trying to access the movie_id because it wasn't passed as a hidden field, so keep in mind that you could've used this route as well, which is also RESTful: "/movies/:movie_id/actors"

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    describe "When i visit a movie page I see the movies info" do
+      it "Also displays all actors in it from youngest to oldest and I see the average age of all the actors" do
+        @movie_1.actors << [@actor_1, @actor_2, @actor_3, @actor_4]
+        visit "/movies/#{@movie_1.id}"
+
+        expect(page).to have_content(@movie_1.title)
+        expect(page).to have_content(@movie_1.creation_year)
+        expect(page).to have_content(@movie_1.genre)
+
+        expect(page).to have_content(@actor_1.name)
```
juliet-e, line 25: You can add `orderly` to the gem file test group as well and use `appear_before` to test order here

spec/features/studios/index_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    describe "When i visit the studip index page I see all the studios and their info" do
+      it "Also displays all their movies and their moveis info" do
+        visit "/studios"
+
+        expect(page).to have_content(@studio_1.name)
+        expect(page).to have_content(@studio_1.location)
+        expect(page).to have_content(@studio_2.name)
+        expect(page).to have_content(@studio_2.location)
+
+        expect(page).to have_content(@movie_1.title)
```
juliet-e, line 25: In your view, you do a great job anchoring these variables to labels  (i.e. `Title:    movie.title`)
Consider making your assertions here more robust by anchoring this data to these same titles. Example:
`expect(page).to have_content("Creation Year: #{@movie_1.creation_year}")`

juliet-e: This will allow you to test that this data is showing up in the right place, rather than randomly floating on the page somewhere

spec/models/movie_spec.rb
```diff
@@ -2,6 +2,41 @@
...
+        expect(@movie_1.youngest_first).to eq([@actor_1, @actor_3, @actor_4, @actor_2])
+      end
+    end
+
+    describe "average_age" do
+      it "should return the actors average age" do
+        @movie_1.actors << [@actor_1, @actor_2, @actor_3, @actor_4]
+        expect(@movie_1.average_age).to eq(18.75)
+      end
+    end
```
juliet-e, line 40: Excellent model testing!

PR #23, opened 2023-10-30
Shawn Carpenter

shawndcarpenter:
I got through  all user stories and the extension. I am proud of how far I got. I was not as confident in user story 2, as I was initially unsure about how to go about calling the actors of a movie. I took a pom and came back and realized it was way easier than i thought and I was, as usual, overthinking things. I am glad that I set up all relationships from the very beginning, including the studio to actors relationship, because it made it easier to get through the extension work later on.


app/views/actor_movies/new.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= form_with url: "/movies/#{@movie.id}", method: :post, local: true do |form| %>
```
cjsim89, line 1: User story 3 states that this form should live on `a movie show page`. I see what you're going for, but it's important to understand that a form can be submitted from _anywhere_ on a website and end up routed _anywhere_ through a given controller. 

That having been said, the setup of your form is great! Just make sure you are conforming to **precisely** what the user story is asking for; no detail in a user story is extraneous or unnecessary!

app/controllers/actor_movies_controller.rb
```diff
@@ -0,0 +1,18 @@
+class ActorMoviesController < ApplicationController
+  def new
+    @movie = Movie.find(params[:id])
```
cjsim89, line 3: This action conforms to the page you created but is not necessary, since we send this data down from the MoviesController to the movie show page as well.

app/controllers/actor_movies_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+    @actors = @movie.actors
+  end
+
+  def create
+    @movie = Movie.find(params[:id])
+    @actors = @movie.actors
+    @actor = @movie.actors.create(actor_params)
```
cjsim89, line 10: Nice job using strong params; don't forget your `create!` bang operator when you use this method - it will be helpful if you need to debug (the error message will just be louder and more in your face.)

app/models/actor.rb
```diff
@@ -0,0 +1,7 @@
+class Actor < ApplicationRecord
+  has_many :actor_movies
+  has_many :movies, through: :actor_movies
+  
+  validates :name, presence: true
+  validates :age, presence: true, numericality: true
```
cjsim89, line 6: Good job creating these validations; note that they will not be necessary on any ICs (but it's good practice)

app/models/movie.rb
```diff
@@ -1,3 +1,16 @@
...
+  has_many :actors, through: :actor_movies
+
+  validates :title, :creation_year, :genre, presence: true
+
+  def youngest_to_oldest
+    actors.order(:age)
+  end
+
+  def average_age
+    actors.average(:age).round(2)
```
cjsim89, line 14: Good AR in your models!

app/views/actor_movies/new.html.erb
```diff
@@ -0,0 +1,8 @@
...
+  <%= form.label :name %>
+  <%= form.text_field :name %>
```
cjsim89, line 3: In addition, the user story also states that we should fill in ONLY the `id` of an actor we want to add to this movie. Granted, this is not a common use case in a "real" application, but again it's important to follow precisely what the user story is asking for.

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,18 @@
...
+    <ul><li><%= actor.age %></li></ul>
+  <% end %>
+</section>
+
+<section class= "Average Age of Actors">
+  <h3>"Average Age of Actors"</h3>
+  <p> <%= @movie.average_age %></p>
+</section>
+
+<p><%= link_to "Add an actor to this movie", "/movies/#{@movie.id}/actors/new" %></p>
```
cjsim89, line 18: Grr, not what the story is asking for (but it does work, so I can give you credit)

spec/features/movies/actors/new_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+RSpec.describe 'new movie actor page' do
+  before :each do
+    @universal = Studio.create!(name: "Universal Studios", location: "Los Angeles")
+    @a24 = Studio.create!(name: "A24", location: "New York")
+    @star_wars = Movie.create!(title: "Star Wars", creation_year: "1979", genre: "sci-fi", studio_id: @universal.id)
+    @lord = Movie.create!(title: "Lord of the Rings", creation_year: "2000", genre: "fantasy", studio_id: @universal.id)
+    @hereditary = Movie.create!(title: "Hereditary", creation_year: "2017", genre: "horror", studio_id: @a24.id)
+    @swiss = Movie.create!(title: "Swiss Army Man", creation_year: "2016", genre: "drama", studio_id: @a24.id)
+    @hereditary.actors.create!([{name: "Toni Collette", age: 42}, {name: "Alex Wolff", age: 27}, {name: "Milly Shapiro", age: 29}])
+    @swiss.actors.create!([{name: "Paul Dano", age: 32}, {name: "Daniel Radcliffe", age: 34}, {name: "Mary Elizabeth Winstead", age: 24}])
```
cjsim89, line 12: Good object setup!

spec/features/movies/actors/new_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    expect(page).to_not have_content("Marika Casteel")
+
+    click_link("Add an actor to this movie")
+    
+    fill_in "name", with: "Marika Casteel"
+    fill_in "age", with: "37"
+
+    expect(page).to have_button
+
+    click_button
```
cjsim89, line 27: Be careful about having a method like this without specifying which button - it will just pick the first button on the page. If you had a story later on that added a 2nd button to the page, this test may fail depending on where that button is located.

cjsim89: In other words, go ahead and always do

```ruby
expect(page).to have_button "Create"
...
click_button "Create"
```

spec/features/movies/actors/new_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    expect(current_path).to eq("/movies/#{@swiss.id}")
+    expect(page).to have_content("Marika Casteel")
+    expect(page).to have_content(37)
+    expect(page).to have_content(31.75)
```
cjsim89, line 32: This is fine, but what value are you testing? Is this just a random number? Your feature tests should have context for the values they are finding, otherwise this could be someone's bank account balance, and your tests would never know 🤷

cjsim89: E.g.

```ruby
expect(page).to have_content("Average age: 31.75")
```
and instantly I know what this number is referring to.

cjsim89: To be even more clear - it's not that the _tests_ need to know, it's that the developer _reading and contributing to_ your tests/application should understand. Little bit of "developer empathy" 👍

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      expect(page).to have_content(@swiss.title)
+      expect(page).to have_content(@swiss.creation_year)
+      expect(page).to have_content(@swiss.genre)
+      expect(page).to_not have_content(@hereditary.title)
+    end
+
+    it "shows movie's actors from youngest to oldest" do
+      visit "/movies/#{@swiss.id}"
+
+      within('section', :text => "Actors from Youngest to Oldest") do
```
cjsim89, line 27: Good use of `within`!

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+        expect("Mary Elizabeth").to appear_before("Paul Dano")
+        expect("Paul Dano").to appear_before("Daniel Radcliffe")
+      end
+    end
+
+    it "show average age of all movie's actors" do
+      visit "/movies/#{@swiss.id}"
+   
+      within('section', :text =>  "Average Age of Actors") do
```
cjsim89, line 36: To make your withins a little bit easier, you can also provide your HTML tags (e.g. `<div>`, `<section>`, etc.) with IDs or Classes to make them easier to find. 
For example if this section had an ID of "average_age", your test could look like: 

```ruby
within("section#average_age") do
   ...
end
```

cjsim89: It ends up being less typing with more precision. What happens if you have two sections on the page with the same text content? Using IDs/classes can help distinguish them faster.

cjsim89 (commented):
Overall great work! You may want to try redoing user story 3's form, just to get the hang of what the story is asking for precisely, but I think your understanding of _how_ to pass that data back and forth is good.

PR #24, opened 2023-10-30
Blaine Kennedy

bkchilidawg:
I think I did well! Only question was on a controller method and if it should've been a model method. The internet said no so I hope it's right!

### DIFF
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -1,0 +1,22 @@
+class MoviesController < ApplicationController
+
+  def show
+    @movie = Movie.find(params[:id])
+    @actors = @movie.actors.order(age: :asc)
+    @average_age = @actors.average(:age)
+  end
+
+  def add_actor
+    @movie = Movie.find(params[:id])
+    actor = Actor.find_by(id: params[:movie][:actor_id]) # Find the actor by ID, returns nil if not found
+
+    if actor
+      @movie.actors << actor
+    else
+      flash[:alert] = "Actor with ID #{params[:movie][:actor_id]} not found."
+    end
+    
+    redirect_to "/movies/#{@movie.id}"
+  end
+end
+
--- a/app/controllers/studios_controller.rb
+++ b/app/controllers/studios_controller.rb
@@ -1,0 +1,6 @@
+class StudiosController < ApplicationController
+  def index
+    @studios = Studio.all
+  end
+end
+
--- a/app/models/actor.rb
+++ b/app/models/actor.rb
@@ -1,0 +1,3 @@
+class Actor < ApplicationRecord
+  has_and_belongs_to_many :movies
+end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -1,3 +1,5 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_and_belongs_to_many :actors
+
 end
--- a/app/views/movies/show.html.erb
+++ b/app/views/movies/show.html.erb
@@ -1,0 +1,30 @@
+<h1><%= @movie.title %></h1>
+<p>Creation Year: <%= @movie.creation_year %></p>
+<p>Genre: <%= @movie.genre %></p>
+
+<br>
+
+<%= form_with(model: @movie, url: url_for(controller: 'movies', action: 'add_actor', id: @movie), method: :post, local: true, turbo: :false) do |form| %>
+  <%= form.label :actor_id, 'Add Actor by ID' %>
+  <%= form.text_field :actor_id %>
+  <%= form.submit 'Add Actor' %>
+<% end %>
+
+<% if flash[:alert] %>
+  <div class="alert alert-danger"><%= flash[:alert] %></div>
+<% end %>
+
+
+<br>
+
+<h2>Actors:</h2>
+<ul>
+  <% @actors.each do |actor| %>
+    <li><%= actor.name %> (Age: <%= actor.age %>)</li>
+  <% end %>
+</ul>
+
+<br>
+
+<p>Average Age of Actors: <%= @average_age %></p>
+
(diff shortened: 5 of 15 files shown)


db/migrate/20231030162554_create_actors_movies.rb
```diff
@@ -0,0 +1,10 @@
+class CreateActorsMovies < ActiveRecord::Migration[7.0]
+  def change
+    create_table :actors_movies do |t|
+      t.references :actor, foreign_key: true
+      t.references :movie, foreign_key: true
+
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 10: Great job setting up your joins table migration!

app/models/actor.rb
```diff
@@ -0,0 +1,3 @@
+class Actor < ApplicationRecord
+  has_and_belongs_to_many :movies
+end
```
abdulredd, line 3: I see that you're using a `has_and_belongs_to_many` association. Is there a reason that you are using this over creating an `ActorMovie` model?

app/models/movie.rb
```diff
@@ -1,3 +1,5 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_and_belongs_to_many :actors
+
```
abdulredd, line 4: I see that you're using a `has_and_belongs_to_many` association. Is there a reason that you are using this over creating an `ActorMovie` model?

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,22 @@
+class MoviesController < ApplicationController
+
+  def show
+    @movie = Movie.find(params[:id])
+    @actors = @movie.actors.order(age: :asc)
+    @average_age = @actors.average(:age)
+  end
```
abdulredd, line 7: We really don't want to be doing any sort of AR within our controller action beyond the basic `find`, `create`, `update`, and `destroy`.

`order` and `average` should be moved to the `Movie` model. Having them in your controller goes against MVC principles.

And with adding these methods to your models, they would then get tested.

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    actor = Actor.find_by(id: params[:movie][:actor_id]) # Find the actor by ID, returns nil if not found
+
+    if actor
+      @movie.actors << actor
+    else
+      flash[:alert] = "Actor with ID #{params[:movie][:actor_id]} not found."
+    end
+    
+    redirect_to "/movies/#{@movie.id}"
+  end
```
abdulredd, line 20: This `add_actor` method should probably be called `create` and this `create` method could be placed in 1 of 2 places; either a `ActorMoviesController` or the `ActorsController` but it most likely will not be placed in the parent controller like you have it here. If you did place it in the child controller, `ActorsController`, you would need to check to see if the `params` included a movie ID before making the join. (If you need help understanding what I mean here, please don't hesitate to reach out)

I think the easiest and most straightforward way to do this would just be in the `ActorMoviesController`. While this isn't an MVC infraction, I think it could have been organized better.

app/controllers/studios_controller.rb
```diff
@@ -0,0 +1,6 @@
+class StudiosController < ApplicationController
+  def index
+    @studios = Studio.all
+  end
+end
```

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,30 @@
...
+<p>Creation Year: <%= @movie.creation_year %></p>
+<p>Genre: <%= @movie.genre %></p>
+
+<br>
+
+<%= form_with(model: @movie, url: url_for(controller: 'movies', action: 'add_actor', id: @movie), method: :post, local: true, turbo: :false) do |form| %>
+  <%= form.label :actor_id, 'Add Actor by ID' %>
+  <%= form.text_field :actor_id %>
+  <%= form.submit 'Add Actor' %>
+<% end %>
```
abdulredd, line 11: This form is unfortunately incorrect in it's implementation. Currently your form is posting to `/movies/:id/add_actor` when it's suppose to be posting to `/movies/:id/actors` since we are creating an association between an actor and a movie.

The updated syntax should look something like this: `<%= form_with url: "/movies/#{@movie.id}/actors, method: :post, local: true, data: { turbo: false } do |form| %>`

In your routes the verb+path would be: `post "/movies/:id/actors", to: "actor_movies#create"`, assuming you have an `ActorMoviesController`, else it would go to `movies#create`. But again, I talked about why it's easier to go to a joins controller before.

config/routes.rb
```diff
@@ -3,4 +3,9 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+  post '/movies/:id/add_actor', to: 'movies#add_actor'
```
abdulredd, line 6: I mentioned this route being incorrect earlier but again this route isn't RESTful. We want the verb+path to be: `post "/movies/:id/actors", to: "actor_movies#create"`

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      expect(page).to have_content(@movie_1.title) 
+      expect(page).to have_content(@movie_1.creation_year) 
+      expect(page).to have_content(@movie_1.genre) 
+      expect(page).to have_content(@actor_1.name) 
+      expect(page).to have_content(@actor_1.age)
+      expect(page).to have_content(@actor_2.name) 
+      expect(page).to have_content(@actor_2.age) 
+      expect(page).to have_content(32) 
+    end
+  end
```
abdulredd, line 33: These tests are ok but I think they could be more verbose, for example right now we are just checking that 32 shows up on the page somewhere. I think it would be much better to have an assertion like this:

`expect(page).to have_content("Average Age: 32") `

This way we reduce the ambiguity of what 32 represents on the page since 32 could be a runtime, and age, the time of day, etc... previously your tests only says does 32 show up anywhere on the page. This could produce a false positive.

PR #25, opened 2023-10-30
Cory Powell

coryrpow:
I did well! Got through all 3 user stories and the extension. I added validations and relationship testing as well and model tested all of the methods I created. Feel really good about this one!

### DIFF
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -1,0 +1,6 @@
+class MoviesController < ApplicationController
+  def show
+    @movie = Movie.find(params[:id])
+  end
+
+end
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -3,4 +3,14 @@ class MoviesController < ApplicationController
     @movie = Movie.find(params[:id])
   end
 
+  def update
+    movie = Movie.find(params[:id])
+    actor = Actor.find(params[:actor_id])
+
+    movie.actors << actor
+    
+    movie.save
+    redirect_to "/movies/#{movie.id}"
+  end
+
 end
--- a/app/models/actor.rb
+++ b/app/models/actor.rb
@@ -1,0 +1,6 @@
+class Actor < ApplicationRecord
+  validates_presence_of :name, :age
+  
+  has_many :movie_actors
+  has_many :movies, through: :movie_actors
+end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -4,4 +4,12 @@ class Movie < ApplicationRecord
   belongs_to :studio
   has_many :movie_actors
   has_many :actors, through: :movie_actors
+
+  def actor_sort
+    actors.order(:age)
+  end
+
+  def actor_age_avg
+    actors.average(:age).round
+  end
 end
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -1,3 +1,7 @@
 class Movie < ApplicationRecord
+  validates_presence_of :title, :creation_year, :genre
+
   belongs_to :studio
+  has_many :movie_actors
+  has_many :actors, through: :movie_actors
 end
--- a/app/models/studio.rb
+++ b/app/models/studio.rb
@@ -2,4 +2,8 @@ class Studio < ApplicationRecord
   has_many :movies
   has_many :movie_actors, through: :movies
   has_many :actors, through: :movie_actors
+
+  def studio_actors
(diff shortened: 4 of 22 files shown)


app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,16 @@
+class MoviesController < ApplicationController
+  def show
+    @movie = Movie.find(params[:id])
+  end
+
+  def update
```
juliet-e, line 6: Like we talked about in office hours, this action would be best reserved for updating a movie's attributes. A more RESTful option would be to route to a MovieActorsController #create action

app/models/actor.rb
```diff
@@ -0,0 +1,6 @@
+class Actor < ApplicationRecord
+  validates_presence_of :name, :age
```
juliet-e, line 2: Nice work validating these attributes! You are not required to create validations during an IC but it's good practice if you choose to include it!

app/models/movie.rb
```diff
@@ -1,3 +1,15 @@
 class Movie < ApplicationRecord
+  validates_presence_of :title, :creation_year, :genre
+
   belongs_to :studio
+  has_many :movie_actors
+  has_many :actors, through: :movie_actors
+
+  def actor_sort
```
juliet-e, line 8: excellent MVC-adherent data operations :)

app/models/studio.rb
```diff
@@ -1,3 +1,11 @@
 class Studio < ApplicationRecord
+  validates_presence_of :name, :location
+
   has_many :movies
+  has_many :movie_actors, through: :movies
+  has_many :actors, through: :movie_actors
```
juliet-e, line 6: This is great! You can also just add one additional association instead of 2 to achieve the same outcome -- because we've already set up the shortcut between movies and actors in the other models, we can just do:
`has_many :actors, through: :movies` without including the joins table

app/models/studio.rb
```diff
@@ -1,3 +1,11 @@
...
+
+  def studio_actors
+    actors.distinct
```
juliet-e, line 9: Excellent work leveraging this additional association for this query

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,21 @@
...
+ 
+<% @movie.actor_sort.each do |actor| %>
+  <li><%= actor.name %>
+  <%= actor.age %></li>
+<% end %>
+
+<p>Average Age of Actors: <%= @movie.actor_age_avg %>
+
+<p> Add an Actor to This Movie:</p>
+<%= form_with url: "/movies/#{@movie.id}", method: :patch, local: true, data: {turbo: false} do |f| %>
```
juliet-e, line 17: As discussed, a more RESTful option could be a POST to "/movies/:id/actors" and have that route to a joins controller

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+      fill_in(:actor_id, with: "#{@actor_4.id}")
+      click_button("Add Actor")
+
+      expect(page).to have_current_path("/movies/#{@movie_1.id}")
+
+      expect(page).to have_content(@actor_4.name)
+
+    end
+  end
+end
```
juliet-e, line 92: Excellent feature testing!

spec/features/studios/index_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+      expect(page).to have_content(@studio_3.name)
+      expect(page).to have_content(@studio_3.location)
+    end
+
+    it "under each studio it shows all the studio's movies
+     with title, creation_year, and genre" do
+      visit "/studios"
+
+      within("#studio_movies-#{@studio_1.id}") do
```
juliet-e, line 40: This successful within block to the rescue today! Huzzah!

spec/models/studio_spec.rb
```diff
@@ -1,7 +1,46 @@
...
+    @movie_actor_4 = MovieActor.create!(movie: @movie_6, actor: @actor_1)
+  end
+
+  describe "#instance _methods" do
+    describe "studio_actors" do
+      it "shows all the actors that are in movies that this studio has made
+      and doesn't show repeat actors" do
+        expect(@studio_1.studio_actors).to eq([@actor_1, @actor_2, @actor_3])
+      end
+    end
```
juliet-e, line 44: Your model tests are also excellent! Just missed the one model spec for MovieActor (2 shoulda matchers assertions for belongs_to)

PR #26, opened 2023-10-30
Kam Kennedy

kameronk92:
I got through user story 3 on this. My internet went out a few times during this IC.. that was frustrating. Overall I think I showed a better understanding of MVC. I still need practice with AR methods to have some more confidence. Was able to rely on what I learned on other projects!


app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,11 @@
+class MoviesController < ApplicationController
+  def show
+    @movie = Movie.find(params[:id])
+  end
+
+  def update
+    @movie = Movie.find(params[:id])
+    @actor = Actor.find(params[:actor_id])
+    @movie.actors << @actor
+  end
```
abdulredd, line 10: You have the right idea here in what you are attempting to do but the only thing that is incorrect is the name of the method and possibly the location of the method as well.

This method should probably be called `create` and this `create` method could be placed in 1 of 2 places; either the `ActorsMoviesController` or the `ActorsController` but not in the `MoviesController` like you have it here. If you did place it in the child controller, `ActorsController`, you would need to check to see if the params included a movie ID before making the join. (If you need help understanding what I mean here, please don't hesitate to reach out)

I think the easiest and most straightforward way to do this would just be in the `ActorsMoviesController`. While this isn't an MVC infraction, I think it could have been organized better.

app/controllers/studios_controller.rb
```diff
@@ -0,0 +1,5 @@
+class StudiosController < ApplicationController
+  def index
+    @studios = Studio.all
+  end
+end
```

app/models/movie.rb
```diff
@@ -1,3 +1,13 @@
...
+  has_many :actor_movies
+  has_many :actors, through: :actor_movies
+
+  def actors_list
+    self.actors.order(age: :asc).pluck(:name)
+  end
+
+  def cast_age
+    Actor.average(:age).to_int
+  end
```
abdulredd, line 12: This method isn't quite correct. In your current implementation, we would be getting the average age of all the actors in the database, and not just the actors of this particular movie. You could pass a parameter of the movie to the `cast_age` method but since we're already in the `Movie` model, that would be overkill.

This is what you want:

```ruby
def cast_age
  actors.average(:age).to_int
end
```

I would also note that some people might consider `to_int` presentation logic since the only reason you are using it is to change how the number you get from the query is displayed. Just something to consider.

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,13 @@
...
+<p>Released in: <%=@movie.creation_year%></p>
+<p>Genre: <%=@movie.genre%></p>
+<p>Actors in this movie: <%=@movie.actors_list.join(', ')%></p>
+<p>Average age of cast members: <%=@movie.cast_age%></p>
+
+<%= form_with url: "/movies/#{@movie.id}", method: :post, data: {turbo: false}, local: true do |form| %>
+ <p> <%= form.label "Add an actor by ID:" %>
+  <%= form.text_field :actor_id %>
+  <%= form.submit "Submit" %></p>
+<% end %>
```
abdulredd, line 13: This form is unfortunately incorrect in it's implementation. Currently your form is posting to `/movies/:id` when it's suppose to be posting to `/movies/:id/actors` since we are creating an association between an actor and a movie.

The updated syntax should look something like this: `<%= form_with url: "/movies/#{@movie.id}/actors, method: :post, data: { turbo: false }, local: true do |form| %>`

In your routes the verb+path would be: `post "/movies/:id/actors", to: "actor_movies#create"`, assuming you have an `ActorMoviesController`, else it would go to `movies#create`. But again, I talked about why it's easier to go to a joins controller before.

app/views/studios/index.html.erb
```diff
@@ -0,0 +1,12 @@
...
+
+<%@studios.each do |s|%>
+  <h3><%=s.name%></h3>
+  <p><%=s.location%></p>
+  <%s.movies.each do |m|%>
+    <h4><%=m.title%></h4>
+    <p><%=m.creation_year%></p>
+    <p><%=m.genre%></p>
+  <%end%>
+<%end%>
```
abdulredd, line 12: Please do not use single variable names in your code. The conventions your learned in Mod 1 still apply in Mod 2.

config/routes.rb
```diff
@@ -3,4 +3,8 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+  get "/studios", to: "studios#index"
+
+  get "/movies/:id", to: "movies#show"
+  post "/movies/:id", to: "movies#update"
```
abdulredd, line 9: I mentioned this route being incorrect earlier but again this route isn't RESTful. We want the verb+path to be: `post "/movies/:id/actors", to: "actor_movies#create"` SInce we are creating an association between an actor and a movie.

A post route would never go to an update action and would also never contain an ending :id attribute in the path.

db/migrate/20231030154156_create_actor_movies.rb
```diff
@@ -0,0 +1,10 @@
+class CreateActorMovies < ActiveRecord::Migration[7.0]
+  def change
+    create_table :actor_movies do |t|
+      t.references :movie, null: false, foreign_key: true
+      t.references :actor, null: false, foreign_key: true
+
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 10: Great job creating the join migration...

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    # And I click submit
+    click_button("Submit")
+    # Then I am redirected back to that movie's show page
+    expect(current_path).to eq("/movies/#{@movie3.id}")
+    # And I see the actor's name is now listed
+    visit "/movies/#{@movie3.id}"
+    expect(page).to have_content("#{@actor3.name}")
+    # (You do not have to test for a sad path, for example if the id submitted is not an existing actor)
+  end
+end
```
abdulredd, line 60: Wow, I really like these tests.... they remove any ambiguity and say explicitly what is being tested. For example, I know just by looking at your tests that 26 is supposed to show up on the page as the average age of cast members. Keep this up, your future teammates will love you for this.

spec/models/movie_spec.rb
```diff
@@ -1,7 +1,36 @@
...
+      expect(@movie3.actors_list).to eq([@actor2.name, @actor1.name])
+    end
+  end
+
+  describe "#cast_age" do
+    it "averages the age of all actors on a given movie" do
+      expect(@movie3.cast_age).to eq(26)
+    end
   end
 end
```
abdulredd, line 36: Great job testing your model methods!

PR #27, opened 2023-10-30
Xander Hendry

XanderHendry:
This IC was a little rough for me. It showed me that I still need to work on staying calm whenever I run into error while on a time limit. I ran into an issue early on (capitalized the 'I' in my index method) and had a bit of a meltdown. but after that I was able to work through the other user stories with relative ease. 
if I wouldn't have spent so much time with that issue I think I would have finished the extension as well.


app/controllers/actor_movies_controller.rb
```diff
@@ -0,0 +1,6 @@
+class ActorMoviesController < ApplicationController
+  def create
+    ActorMovie.create!({ movie_id: params[:id], actor_id: params[:actor_id]})
+    redirect_to "/movies/#{params[:id]}"
+  end
+end
```
abdulredd, line 6: Great job using your joins controller to make the association between a movie and an actor!

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,6 @@
+class MoviesController < ApplicationController
+  def show
+    @movie = Movie.find(params[:id])
+    @actors = @movie.actors
+  end
+end
```
abdulredd, line 6: It's perfectly fine to call on `@movie.actors` from within the Movie show view 

What you have here isn't wrong either :) 

I will point out, however, that I think you are not ordering the actors from youngest to oldest as specified in US2.

app/controllers/studios_controller.rb
```diff
@@ -0,0 +1,10 @@
+class StudiosController < ApplicationController
+  def index
+    @studios = Studio.all
+  end
+
+  def show
+    @studio = Studio.find(params[:id])
+    @studio_actors = @studio.studio_actors
+  end
+end
```
abdulredd, line 10: Similar to what I mentioned earlier in your `MoviesController`, it's perfectly fine to do `@studio.studio_actors` from within the view. It may seem like an infraction to do it that way, but you aren't really calling on AR, but more so on a model method.

Again, doing it in the controller isn't incorrect either.

app/models/movie.rb
```diff
@@ -1,3 +1,9 @@
 class Movie < ApplicationRecord
   belongs_to :studio
+  has_many :actor_movies
+  has_many :actors, through: :actor_movies
+
+  def average_actor_age
+    actors.pluck(:age).sum / actors.count
+  end
```
abdulredd, line 8: While in theory you are using active record here, there is an average aggregate function that does all of this work for you. Just in case, in SQL there are 5 aggregate functions: COUNT, SUM, AVG, MIN, and MAX

Here's the updated code:

```ruby
def average_actor_age
  actors.average(:age)
end
```

app/models/studio.rb
```diff
@@ -1,3 +1,7 @@
 class Studio < ApplicationRecord
   has_many :movies
+
+  def studio_actors
+    Movie.joins(:actors).where("movies.studio_id = #{self.id}")
+  end
```
abdulredd, line 6: Great job working through the join. If I'm not mistaken though, this query would most likely return you Movie objects when in fact what you want is Actor objects.

There are 2 ways to do this:

Using pure AR:  `Actor.joins(:movies).where("movies.studio_id = #{self.id}").distinct`

Setting up an additional association in your Studio model:

```ruby
has_many :actors, through: :movies
# This one might be a little tough to understand since we are not technically using a joins table but if you diagram the schema, yo uwill see that from the Studios table, we can get to actors through the Movies table because it has access to the Actors table already through the ActorMovies table.
```

Then you can just have this logic in your `studio_actors` method: `actors.distinct`

If you need help understanding the last one, please do not hesitate to reach out.

app/views/movies/show.html.erb
```diff
@@ -0,0 +1,16 @@
...
+<p><%= actor.name %>  Age: <%= actor.age %></p>
+<% end %>
+<p>Average Cast Age: <%= @movie.average_actor_age %></p>
+
+<section id="add_actor">
+  <%= form_with url: "/movies/#{@movie.id}/actor_movies", method: :patch, local: true do |f| %>
+  <%= f.label :actor_id, "Actor ID" %>
+  <%= f.text_field :actor_id %>
+  <%= f.submit "Submit" %>
+  <% end %>
```
abdulredd, line 15: This form is so very close but unfortunately incorrect in its current implementation. Currently, your form is making a patch request to `/movies/:movie_id/actor_movies` when it's supposed to be making a post request to `/movies/:id/actors` since we are creating an association between an actor and a movie.

I do understand your use of `actor_movies` but if you wanted to use this route, it would be something like: `url: "/actor_movies", method: :post` in order to keep it RESTful. But you would then need to pass the movie ID as additional params. It gets messy quickly.

The updated syntax should look something like this: `<%= form_with url: "/movies/#{@movie.id}/actors, method: :post, local: true do |f| %>`

In your routes the verb+path would be: `post "/movies/:movie_id/actors", to: "actor_movies#create"`, assuming you have an `ActorMoviesController`.

app/views/studios/index.html.erb
```diff
@@ -0,0 +1,12 @@
...
+<% @studios.each_with_index do |studio, index| %>
+  <h3><%= studio.name %></h3>
+  <p>Location: <%= studio.location %></p>
+  <p>Film Library</p>
+  <% studio.movies.each do |movie| %>
+      <h6><%= movie.title %></h6>
+      <p>Creation Year: <%= movie.creation_year %></p>
+      <p>Genre: <%= movie.genre %></p>
+  <% end %>
+<% end %>
```
abdulredd, line 12: Interesting that you went with `each_with_index` but never used the `index` block variable. I think it might be better to just use `each` next time.

config/routes.rb
```diff
@@ -3,4 +3,10 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/studios", to: "studios#index"
+  get "/studios/:id", to: "studios#show"
+  get "/movies/:id", to: "movies#show"
+  patch "/movies/:id/actor_movies", to: "actor_movies#create"
```
abdulredd, line 10: I mentioned this route being incorrect earlier but again this route isn't RESTful. We want the verb+path to be: `post "/movies/:id/actors", to: "actor_movies#create"` SInce we are creating an association between an actor and a movie. A `patch` wouldn't be used in combination with a `create` action.

db/migrate/20231030165851_create_actor_movies.rb
```diff
@@ -0,0 +1,10 @@
+class CreateActorMovies < ActiveRecord::Migration[7.0]
+  def change
+    create_table :actor_movies do |t|
+      t.references :movie, null: false, foreign_key: true
+      t.references :actor, null: false, foreign_key: true
+
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 10: Great job creating the join table!

spec/features/movies/show_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+        fill_in :actor_id, with: @miles.id
+        click_button "Submit"
+        end
+
+        expect(current_path).to eq("/movies/#{@tg_maveric.id}")
+        expect(page).to have_content(@miles.name)
+      end
+    end
+  end
+end
```
abdulredd, line 62: Great job using `within` blocks as part of your tests. They really help eliminate ambiguity and help with test organization.

spec/features/studios/index_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        expect(page).to have_content("Creation Year: #{@jurassic_world.creation_year}")
+        expect(page).to have_content("Genre: #{@jurassic_world.genre}")
+        expect(page).to have_content(@mario.title)
+        expect(page).to have_content("Creation Year: #{@mario.creation_year}")
+        expect(page).to have_content("Genre: #{@mario.genre}")
+      end
+  
+    end
+  end
+end
```
abdulredd, line 56: Wow, I really like these tests.... they remove any ambiguity and say explicitly what is being tested. For example, I know just by looking at your tests that 1997 is supposed to show up on the page as the year for a movie and not as part of some phone number. Keep this up, your future teammates will love you for this.

app/models/movie.rb
```diff
@@ -1,3 +1,9 @@
...
 end
```
abdulredd, line 9: I think your model is missing the method to order the actors by age.

spec/models/movie_spec.rb
```diff
@@ -3,5 +3,26 @@
...
+    ActorMovie.create!({ movie_id: @tg_maveric.id, actor_id: @tom.id})
+    ActorMovie.create!({ movie_id: @tg_maveric.id, actor_id: @jennifer.id}) 
+   end
+
+  describe ".average_actor_age" do 
+    it "returns the average age of all Actors in a Movie" do 
+      expect(@tg_maveric.average_actor_age).to eq(56)
+    end
   end
 end
```
abdulredd, line 28: Missing actor ordering method and therefore the accompanying tests.

